PR #4, opened 2016-07-20
Refactor: move pluginRegistry in private area

just-boris:
* all core components should import `./util/pluginsRegistry` to read all registered plugins
* `pluginsApi` is now file only for export public apis (as it should be by its name)
* plugins from `plugins` folder now also use global `allure.api` to register


baev:
thanks!

PR #16, opened 2016-11-19
Rft steps [fixes #19]

### LINKED ISSUE #19: Add support for test fixtures

baev:
screen shot 2016-11-19 at 23 13 08


face/src/helpers/or.js
```diff
@@ -1,3 +1,4 @@
-export default function(a, b) {
-    return a || b;
+export default function(...args) {
+    args.pop();
+    return args.reduce((a, b) => !!(a || b), false);
```
just-boris: no need to make this so complex.
When you have more than 2 operands, you should move this logic somewhere else.

face/src/components/execution/ExecutionView.hbs
```diff
@@ -0,0 +1,8 @@
+{{#if (or before.length test.length after.length)}}
```
just-boris: rather than compute it here, add an extra return value into `serializeData`

face/src/components/execution/stage.hbs
```diff
@@ -0,0 +1,19 @@
+{{#if name}}
```
just-boris: this partial is used only once. Would be better to inline this into `stages-block`

face/src/helpers/or.js
just-boris: Handlebars by design intended not to have logic inside.
Actually, the `or` helper is an anti-pattern. While it is quite okay to break the rule for two values, it would be dangerous to make it more complex

baev: I got it

face/src/components/execution/ExecutionView.js
```diff
@@ -26,6 +28,17 @@ class ExecutionView extends View {
...
+
+    @on('click .attachment-row')
+    onAttachmentClick(e) {
+        const attachmentUid = $(e.currentTarget).data('uid');
+        router.toUrl(this.options.baseUrl + '/' + attachmentUid);
+    }
+
+    @on('click .attachment-row__link')
+    onAttachmentFileClick(e) {
+        e.stopPropagation();
```
just-boris: why do you need this?

face/src/components/execution/ExecutionView.js
```diff
@@ -26,6 +28,17 @@ class ExecutionView extends View {
     onStepClick(e) {
         this.$(e.currentTarget).parent().toggleClass('step_expanded');
     }
+
+    @on('click .attachment-row')
+    onAttachmentClick(e) {
+        const attachmentUid = $(e.currentTarget).data('uid');
```
just-boris: `this.$(e.currentTarget)` and you don't need to import `jquery`

face/src/data/testcase/TestcaseModel.js
```diff
@@ -11,16 +12,15 @@ export default class TestcaseModel extends Model {
...
+                .concat(makeArray(attachments));
         }
-        function collectAttachments(steps) {
-            return orEmpty(steps).reduce((attachments, step) => {
-                return attachments.concat(collectAttachments(step.steps), orEmpty(step.attachments));
-            }, []);
-        }
-        this.allAttachments = collectAttachments(this.get('steps'))
-            .concat(orEmpty(this.get('attachments')));
+        this.allAttachments = makeArray(this.get('beforeStages'))
```
just-boris: You are always wrapping this value into `makeArray`. Would be better to do it only once in `model.parse` and then always trust that value.

baev: I don't feel like make it perfect right now, and I guess that is fine at the point, probably will fix it later (that change is only the beginning)

face/src/components/execution/ExecutionView.js
just-boris: There is a difference between `.attachment-row` and `.attachment-row__link`. Attachment-link has `target="_blank"` and will be opened in a new tab. No need to prevent this.

face/src/data/testcase/TestcaseModel.js
just-boris: Okay. I have let you know about possible improvement.
If you would like, you can leave as it is now.

PR #22, opened 2016-11-21
Display full name for test cases [fixes #17]

### LINKED ISSUE #17: Add class name to test cases

baev:
screen shot 2016-11-21 at 18 40 02

### DIFF
--- a/face/src/components/testcase/TestcaseView.hbs
+++ b/face/src/components/testcase/TestcaseView.hbs
@@ -1,4 +1,5 @@
 <a class="pane__overlay" href="#{{route.baseUrl}}/{{uid}}"></a>
+<div class="pane__subtitle long-line line-ellipsis">{{fullName}}</div>
 <h2 class="pane__title">
     <span class="label label_status_{{status}}">{{status}}</span>
     <span class="pane__controls">
--- a/face/src/components/pane-set/styles.css
+++ b/face/src/components/pane-set/styles.css
@@ -27,6 +27,11 @@
       background: $pane-overlay-bg-hover;
     }
   }
+  &__subtitle {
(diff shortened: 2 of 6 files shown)


baev:
Summon @eroshenkoam @just-boris

just-boris:
This line is going to jump when you are switching between cases that have a full name to cases, that have not. Let's reserve some height for this block, or put it bellow the usual name.

baev:
> This line is going to jump when you are switching between cases that have a full name to cases

Is that a problem? Usually all the test cases have a full name, so everything should be fine. Some extra reserved space looks bad imho. 

Probably we should aways use only one line and hide the rest of it

just-boris:
> Is that a problem? 
I was getting some complaints about that in Allure 1 and did some work to improve it.
Here the story comes again.

Of course we can postpone the fix, until somebody else will report it.

face/src/components/testcase/TestcaseView.hbs
```diff
@@ -1,4 +1,7 @@
 <a class="pane__overlay" href="#{{route.baseUrl}}/{{uid}}"></a>
+{{#if fullName}}
+    <div class="pane__subtitle">{{fullName}}</div>
+{{/if}}
```
just-boris: empty div has zero height, you can safely get rid of #if to make it cleaner

baev:
That is so strange to have so much reserved space on the top of the screen...

screen shot 2016-11-21 at 19 03 21

just-boris:
Agree.
What about the second option: put full name bellow the title?

baev:
screen shot 2016-11-21 at 19 11 15

baev:
test this please

PR #72, opened 2017-01-23
handle null id [fix #71]

### LINKED ISSUE #71: Skip history if test case id is null

ehborisov:


allure-generator/src/main/java/io/qameta/allure/allure2/Allure2ResultsReader.java
```diff
@@ -75,7 +77,12 @@ public Allure2ResultsReader(AttachmentsStorage storage) {
                     TestCaseResult dest = new TestCaseResult();
                     dest.setUid(generateUid());
 
-                    dest.setId(result.getId());
+                    String id = Optional.ofNullable(result.getId()).orElseGet(() -> {
```
baev: What is the point of generation random id here?

allure-generator/src/main/java/io/qameta/allure/history/HistoryResultAggregator.java
```diff
@@ -35,22 +35,24 @@
...
-                    .withStatus(result.getStatus())
-                    .withStatusDetails(Objects.isNull(result.getFailure()) ? null : result.getFailure().getMessage())
-                    .withTime(result.getTime())
-                    .withTestRunName(testRun.getName());
-
-            List<HistoryItem> newItems = Stream.concat(Stream.of(newItem), data.getItems().stream())
-                    .limit(5)
-                    .collect(Collectors.toList());
-            data.setItems(newItems);
+            if(Objects.nonNull(result.getId())){
```
baev: to reduce complexity it is better to use return if id is null

allure-generator/src/main/java/io/qameta/allure/allure2/Allure2ResultsReader.java
ehborisov: This is where test case result with null id is being parsed. We discussed that we wand to generate an id if it has not been provided, thus making this test an orphan, aren't we?

baev: > We discussed that we wand to generate an id if it has not been provided

Nope, that is gonna be a feature - use null id to disable history & statistics for test case

ehborisov: Ok, I made the corrections

baev:
@ehborisov lets also update this line 
https://github.com/allure-framework/allure2/blob/master/allure-generator/src/main/java/io/qameta/allure/ProcessStage.java#L94

PR #99, opened 2017-02-02
Configure cache for npm task (fixes #95)

### LINKED ISSUE #95: Add cache for npm gradle task

baev:


allure-web/build.gradle
```diff
@@ -5,12 +5,25 @@ apply from: "${gradleScriptDir}/bintray.gradle"
...
     download = true
 }
 
-processResources.dependsOn npm_run_build
+task buildWeb(type: NpmTask) {
+    inputs.files(fileTree('node_modules'))
+    inputs.files(fileTree('src'))
+    inputs.file('package.json')
+    inputs.file('jasmine.json')
+    inputs.file('webpack.config.js')
```
just-boris: What does this file list do?
Should we also add `.eslintrc` and `.babelrc` files there? Build will not work without them.

baev: gradle will skip the task in case there are no changes in these files

PR #126, opened 2017-02-09
Support attaching tsv files (fixes #123)

sseliverstov:
+ add proper icon for video
+ add icon for uri list
+ image file icon for sag
+ test data


allure-web/src/util/attachmentType.js
```diff
@@ -16,22 +21,62 @@ export default function typeByMime(type) {
...
+                                .map(line => line.trim())
+                                .filter(line => line.length > 0)
+                                .map(line => ({
+                                    comment: line.startsWith('#'),
+                                    text: line
+                                }))
+            };
         default:
+            return {
+                type: '*',
```
just-boris: how about returning `type: null` when it is unknown?

Makes more sense, so you can write something like `if(!type) {...}` later to handle case when type is not defined.

sseliverstov: fixed, 10x!

PR #167, opened 2017-02-15
Hide copy to clipboard icon if fullName empty (fixes #158)

### LINKED ISSUE #158: Hide copy to clipboard icon in case of empty full name
screen shot 2017-02-14 at 19 14 59

sseliverstov:


allure-web/src/components/testcase/TestcaseView.hbs
```diff
@@ -1,7 +1,9 @@
 <a class="pane__overlay" href="#testcase/{{uid}}"></a>
-<div class="pane__subtitle long-line line-ellipsis">
-    <span class="fa fa-files-o fullname__copy" data-tooltip="Copy to clipboard"></span>
-    <span class="fullname__body">{{fullName}}</span>
+{{#if (is-def fullName)}}
```
baev: You don't need `is-def` helper here. It is only needed to split `undefined` with `0` (for numbers, not strings):

```js
export default function isDef(variable) {
    return (variable || typeof variable === 'number');
}
```

just-boris:
@baev 
https://pixabay.com/en/animals-nature-life-hand-yellow-71895/

baev:
ok, thanks

allure-web/src/components/testcase/TestcaseView.hbs
sseliverstov: done

PR #195, opened 2017-03-01
default statuses & fix for settings (fixes #181)

### LINKED ISSUE #181: Results with statuses 'skipped' and 'unknown' are not shown on xunit tab by default
image

all statuses should be selected by default

ehborisov:


just-boris:
Actually, the problem was that I have applied plugins in a custom order.
I have added a change that makes it right.

But also it makes sense not to deal with badly documented stuff in favor of explicit coding.

allure-web/src/data/settings/SettingsModel.js
```diff
@@ -1,24 +1,24 @@
...
+            sidebarCollapsed: false,
+            visibleStatuses: {
+                failed: true,
+                broken: true,
+                skipped: true,
+                unknown: true,
+                passed: true
+            },
+            showGroupInfo: true
+        }
```
just-boris: missing semicolon

allure-web/src/data/settings/SettingsModel.js
```diff
@@ -1,24 +1,24 @@
-import {Model} from 'backbone';
-import {protoprop} from '../../decorators';
+import {Model} from "backbone";
```
just-boris: Strings must use singlequote

PR #198, opened 2017-03-03
fix web tests configuration

just-boris:
* `test` folder should be part of the input
* simplify npm test command


just-boris:
Now it runs fine

image

FYI @ehborisov

baev:
@just-boris thanks!

PR #199, opened 2017-03-04
Use Jest as a frontend testing tool (closes #189)

### LINKED ISSUE #189: Replace Jasmine with Jest
Jest includes some nice defaults, that now we are configuring manually: https://github.com/allure-framework/allure2/blob/master/allure-web/test/bootstrap.js

Test codebase will not be changed, because Jest and Jasmine has the same API

just-boris:


just-boris:
Finally, the coverage report works. Numbers are not good, though:

image

just-boris:
FYI @cpojer, there is an example of Jest setup for Backbone+Handlebars stack.

baev:
@just-boris thanks!

cpojer:
@just-boris that's awesome. Can I encourage you to write a blog post about all of this? :)

just-boris:
Maybe. Here I have implemented a Handlebars transformer. It can be extracted as a separate package and then presented via blog post.

PR #200, opened 2017-03-06
separate settings for tabs (fixes #191)

### LINKED ISSUE #191: Separate filters and configuration for each tab

ehborisov:


allure-web/src/components/testcase-table/TestcaseTableView.js
```diff
@@ -13,17 +13,21 @@ class TestcaseTableView extends DataGridView {
     settingsKey = 'testCaseSorting';
 
     initialize() {
-        this.listenTo(settings, 'change:visibleStatuses', this.render);
+        this.statusesKey = 'testcaseTableStatuses';
```
just-boris: this may be just a constant on top of the file

allure-web/src/components/testcase-table/TestcaseTableView.js
```diff
@@ -13,17 +13,21 @@ class TestcaseTableView extends DataGridView {
...
 
     onRender() {
         // this.highlightItem(this.options.currentCase);
-        this.showChildView('statuses', new StatusToggleView());
+        this.showChildView('statuses', new StatusToggleView({statusesKey: this.statusesKey}));
     }
 
     serializeData() {
-        const statuses = settings.get('visibleStatuses');
+        const statuses = settings.get(this.statusesKey);
```
just-boris: Better to put a condition here
```
const statuses = settings.get(SETTINGS_KEY) || settings.get('visibleStatuses');
```
then do a force-update in the `initialize`

allure-web/src/components/tree/TreeView.js
```diff
@@ -61,7 +65,7 @@ class TreeView extends View {
     }
 
     serializeData() {
-        const statuses = settings.get('visibleStatuses');
+        const statuses = settings.get(this.statusesKey);
```
just-boris: Same as in `TestcaseTableView.js`

allure-web/src/helpers/statistic-bar.js
```diff
@@ -2,8 +2,9 @@ import {SafeString} from 'handlebars/runtime';
 import settings from '../util/settings';
 import {values} from '../util/statuses';
 
-export default function statisticBar(statistic) {
-    const statuses = settings.get('visibleStatuses');
+export default function (statistic, tabName){
+    const statusesKey = tabName + 'visibleStatuses';
```
just-boris: What is there?
In all other files you use `tabName`  without suffix

ehborisov: no, I don't.
see TreeView.js:17

allure-web/src/components/testcase-table/TestcaseTableView.js
ehborisov: I don't get it, sorry. Isn't `initialize` being called first, so the `settings.save(this.statusesKey, ...)` should be exactly there?

just-boris: No, I mean that you should not save anything here.
You can just read settings from the old and new place.

allure-web/src/components/tree/TreeView.js
```diff
@@ -14,9 +14,13 @@ class TreeView extends View {
         this.state = state;
         this.baseUrl = baseUrl;
         this.tabName = tabName;
-        this.statusesSelect = new StatusToggleView();
+        this.statusesKey = tabName + 'visibleStatuses';
```
just-boris: Maybe it would be better to make this key with dot, like `tabName + '.visibleStatuses'`.
Not a big change, but better to read in debug

allure-web/src/components/status-toggle/StatusToggleView.js
```diff
@@ -40,7 +40,7 @@ class StatusToggleView extends PopoverView {
     }
 
     serializeData() {
-        const statuses = settings.get(this.statusesKey);
+        const statuses = settings.get(this.statusesKey) || settings.get('visibleStatuses');
```
just-boris: Well, we have repeated that code so many times.
Maybe it would be better to create a new method `getVisibleStatuses(statusesKey)` in the `SettingsModel`?

ehborisov: done

baev:
@ehborisov thanks!

PR #201, opened 2017-03-07
setup yarn (fixes #97)

### LINKED ISSUE #97: Migrate to yarn

just-boris:
Build time before: ~5min
Build time after: ~5min

### DIFF
--- a/allure-web/.gitignore
+++ b/allure-web/.gitignore
@@ -3,3 +3,4 @@ coverage
 etc
 node_modules
 node
+yarn.lock
--- a/allure-web/build.gradle
+++ b/allure-web/build.gradle
@@ -6,11 +6,11 @@ apply plugin: 'maven'
 
 node {
     version = '6.3.1'
-    npmVersion = '3.10.3'
+    yarnVersion = '0.20.3'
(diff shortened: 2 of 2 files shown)


just-boris:
To be precise, I have run build 3 times. Result is better with every attempt:
https://ci.qameta.in/job/allure2_pull-request/139/ – 5 min 28 sec
https://ci.qameta.in/job/allure2_pull-request/141/ – 4 min 49 sec
https://ci.qameta.in/job/allure2_pull-request/142/ – 4 min 30 sec

baev:
@just-boris thanks!

allure-web/.gitignore
```diff
@@ -3,3 +3,4 @@ coverage
 etc
 node_modules
 node
+yarn.lock
```
lanwen: you should commit this file as mentioned in https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control

just-boris: I have already tried that in other projects. It causes merge conflicts for no reason. This problem is bigger than benefits from fixed package versions

PR #204, opened 2017-03-13
Colored arrow for fixtures and nested steps (fixes #203)

### LINKED ISSUE #203: Colorize arrows with max status

sseliverstov:


allure-web/src/helpers/arrow.js
```diff
@@ -0,0 +1,11 @@
+import {SafeString} from 'handlebars/runtime';
+import {values} from '../util/statuses';
+
+
+export default function(item) {
+    var status = values.slice().reverse().reduce((result, current) => {
+        return item && (item[current] || item.status === current) ? current : result;
+    });
```
just-boris: Do not understand how it works. The given object

```
{
  failed: 2,
  passed: 1
}
```
Will give you `passed` because you are getting whatever comes last, regardless actual value.

sseliverstov: reversed values:
'unknown', 'skipped', 'passed', 'broken', 'failed'
So init value for reduce is 'unknown'
1) for 'undefined' instead item - result is initial item - 'unknown'
2) for object with key 'status' - result will be 'value' of key 'status'
3) finally if item is statistics object, we take last status with non zero value in reversed 'values' order. For you example it is failed

We play with colored arrows for test groups, but decide that it is hard to understand.
Now this helper use only 'status' key. 

I fix it in next commit

allure-web/src/components/execution/stages-block.hbs
```diff
@@ -1,15 +1,19 @@
...
             {{name}}
         </div>
         <div class="{{b 'step' 'content'}}">
             {{#each stages}}
                 {{#if name}}
                     <div class="{{b 'step'}}">
                         <div class="{{b 'step' 'title' hasContent=(or steps.length attachments.length) stage=true}}">
-                            <span class="step__arrow block__arrow"></span>
+                            {{#if (or steps.length attachments.length)}}
+                                <span class="step__arrow block__arrow">{{arrow this}}</span>
```
just-boris: How about passing `status` here directly?

```
{{arrow status}}
```

allure-web/src/helpers/arrow.js
```diff
@@ -0,0 +1,8 @@
+import {SafeString} from 'handlebars/runtime';
+import {values} from '../util/statuses';
+
+
+export default function(item) {
+    var status = (item && item.status) ? item.status : 'unknown';
```
just-boris: then you do not do a check here, you can just do 

```js
export default function (status = 'unknown') {
  return new SafeString(`<span class="fa fa-chevron-right fa-fw text_status_${status}"></span>`);
}
```

PR #207, opened 2017-03-15
Time for steps and attachmets/substeps stats on parent level (fixes #83)

### LINKED ISSUE #83: Display time for steps

sseliverstov:


allure-web/src/components/execution/ExecutionView.js
```diff
@@ -29,10 +29,26 @@ class ExecutionView extends View {
         attachmentEl.parents('.step').addClass('step_expanded');
     }
 
+    calculateSteps(steps, key) {
+        return steps.map(step => {
+            var steps = this.calculateSteps(step.steps, key);
+            step[key + 'Count'] = step[key].length + steps.reduce((count, cur) => { return count + cur[key + 'Count'];}, 0);
+            step.steps = steps;
```
just-boris: is this a reference to the parent?
Would it be better to call it `parent` then?

sseliverstov: fixed

allure-web/src/components/execution/ExecutionView.js
```diff
@@ -29,10 +29,26 @@ class ExecutionView extends View {
         attachmentEl.parents('.step').addClass('step_expanded');
     }
 
+    calculateSteps(steps, key) {
+        return steps.map(step => {
+            var children = this.calculateSteps(step.steps, key);
+            step[key + 'Count'] = step[key].length + children.reduce((count, cur) => { return count + cur[key + 'Count'];}, 0);
+            step.steps = children;
```
just-boris: Do we actually need an assignment there?
It is almost the same value as it was before on the top line.

just-boris: If I was working on this, I probably would write something like:

```js
calculateSteps(steps, key) {
  const countKey = `${key}Count`;
  steps.forEach(step => {
    this.calculateSteps(step.steps, key);
    step[countKey] = step[key].length + step.steps.reduce(
      (count, cur) => count + cur[countKey],
      0
    );
  });
}
```

sseliverstov: Add your pretty code ;) Thank you.

sseliverstov:
@baev =)

baev:
@sseliverstov thanks!

PR #247, opened 2017-04-10
screen-diff plugin updates

just-boris:
* Fixed problems with resize
* Roll back ES6 stuff. For plugins we do not have Babel therefore we should write code compatible with all browsers.
* Screen-diff test data is now placed as separate dataset, in order not to litter main development environment.


just-boris:
My plugin development setup, if you need so: 
1) Build allure-commandline via `./ gradlew build`
2) Symlink plugin folder into commandline distribution: 
```
cd allure-commandline/build/install/allure/plugins/
ln -s ../../../../../plugins/screen-diff-plugin/src/dist screen-diff-plugin
```
3) Generate report `./allure-commandline/build/install/allure/bin/allure generate -c allure-generator/test-data/screen-diff`
4) Whenever you made some changes in the plugin, you need to run step 3 again.

PR #250, opened 2017-04-10
use ABC sorting order for tree-view items

just-boris:
Currently it shows all items like
* Z
* Y
* X

This neither not usable or understandable. Before we will implement fully-featured sorting, we need to fix at least in this way.

### DIFF
--- a/allure-generator/src/test/javascript/stubs/handlebars-compiler.js
+++ b/allure-generator/src/test/javascript/stubs/handlebars-compiler.js
@@ -5,8 +5,6 @@ const path = require('path');
 const fs = new MemoryFS();
 const file = process.argv[2];
 
-console.log('SETUP HOHOHO');
-
 const compiler = webpack({
     entry: file,
     output: {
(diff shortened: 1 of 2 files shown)


allure-generator/src/test/javascript/stubs/handlebars-compiler.js
```diff
@@ -5,8 +5,6 @@ const path = require('path');
 const fs = new MemoryFS();
 const file = process.argv[2];
 
-console.log('SETUP HOHOHO');
-
```
just-boris: This line used to fail tests. Sure, we need to get rid of this and enable JS-tests in the build again.

baev: yep, that is my fault. I can't get them working again after merging generator & web together. Could you please take a look?

baev:
thanks @just-boris

allure-generator/src/test/javascript/stubs/handlebars-compiler.js
just-boris: After this change, I run `yarn test` and it passed localy.
How else should I test?

PR #253, opened 2017-04-11
Add an ability to view results from history (fixes #69)

### LINKED ISSUE #69: Add an ability to view test case from history
You need to be able to view test case in separate pane when click to related history item

baev:


allure-generator/src/main/javascript/helpers/timeago.js
```diff
@@ -0,0 +1,8 @@
+import moment from 'moment';
+import runtime from 'injectify/runtime';
```
just-boris: Replace this with `handlebars/runtime`.

Injectify is a kind of semi-internal Yandex thing, we doin't need this

allure-generator/src/main/javascript/helpers/timeago.js
```diff
@@ -0,0 +1,8 @@
+import moment from 'moment';
```
just-boris: Also I would suggest using `date-fns` instead of `moment`.
https://date-fns.org/docs/distanceInWordsToNow

The API is much simpler

```js
import distanceInWordsToNow from 'date-fns/distance_in_words_to_now';
import format from 'date-fns/format';

const threshold = 24 * 3600 * 1000;

export default function (date) {
     const dateString = Date.now() - date < threshold ? distanceInWordsToNow(date) : format(date, 'DD MMMM YYYY');
     return new runtime.SafeString(`<span title="${format('DD MMMM YYYY, H:mm:ss')}">${dateString}</span>`);
 } 
```

baev:
@qameta-ci retest this please

PR #256, opened 2017-04-12
Rft test case page

baev:
screen shot 2017-04-12 at 20 03 45


allure-generator/src/main/javascript/plugins/testcase-severity/SeverityView.hbs
```diff
@@ -0,0 +1,3 @@
+{{#if severity}}
+    <span>Severity: {{severity}}</span>
+{{/if}}
```
just-boris: Do you like to have so many redundant files? Instead of this, you can do inline template


```js
import {escapeExpression} from 'handlebars/runtime';

class SeverityView extends View {
   //..skipped

   template(data) {
     return severity ? `Severity ${escapeExpression(data.severity)}`;
   }
}

```

allure-generator/test-data/demo/d3df6bc4-475d-436b-b99f-3a4eaba6d1a7-testsuite.xml
```diff
@@ -2,7 +2,7 @@
 <ns2:test-suite xmlns:ns2="urn:model.allure.qatools.yandex.ru" start="1391003394188" stop="1391003394494">
     <name>my.company.AllureMigrationTest</name>
     <test-cases>
-        <test-case start="1412949539363" stop="1412949993580" status="skipped" severity="normal">
+        <test-case status="skipped" severity="normal">
```
just-boris: how about handling the well known case `<test-case stop="">`?

baev: this will require more changes in parser, maybe later

PR #268, opened 2017-04-16
fix npe in commandline (fixes #262)

### LINKED ISSUE #262: Commandline called without args throws NPE
Exception in thread "main" java.lang.NullPointerException
        at io.qameta.allure.CommandLine.run(CommandLine.java:119)
        at java.util.Optional.orElseGet(Optional.java:267)
        at io.qameta.allure.CommandLine.main(CommandLine.java:72)

ehborisov:


allure-commandline/src/main/java/io/qameta/allure/CommandLine.java
```diff
@@ -116,6 +116,9 @@ public ExitCode run() {
         }
 
         final String parsedCommand = commander.getParsedCommand();
+        if (Objects.isNull(parsedCommand)){
+            return ExitCode.ARGUMENT_PARSING_ERROR;
```
baev: lets print usage at this point

baev:
try

```
$ allure -q
```

allure-commandline/src/test/java/io/qameta/allure/CommandLineTest.java
```diff
@@ -224,4 +224,15 @@ public void shouldPrintPluginList() throws Exception {
         assertThat(run)
                 .isEqualTo(NO_ERROR);
     }
+
+    @Test
+    public void shouldHandleUnknownOptions() {
```
baev: `-q` is not an unknown option by the way. See https://github.com/allure-framework/allure2/blob/0d1517e4d44d30812ea38b75d82ac27f7343712c/allure-commandline/src/main/java/io/qameta/allure/option/VerboseOptions.java

Unknown options works fine:

```bash
$ allure --unknown

Could not parse arguments: Expected a command, got --unknown
Usage: allure [options] [command] [command options]
  Options:
    --help
      Print commandline help.
    -q, --quiet
      Switch on the quiet mode.
      Default: false
    -v, --verbose
      Switch on the verbose mode.
      Default: false
    --version
      Print commandline version.
      Default: false
```

allure-commandline/src/main/java/io/qameta/allure/CommandLine.java
```diff
@@ -75,6 +75,10 @@ public static void main(final String[] args) throws InterruptedException {
 
     @SuppressWarnings("PMD.AvoidLiteralsInIfCondition")
     public Optional<ExitCode> parse(final String... args) {
+        if (args.length == 0) {
```
baev: I guess no need to check arguments length at this point. `ParameterException` is thrown for all the unknown options and commands, am I right?

ehborisov: Technically, yes, but then the test shouldParseEmptyArguments checks nothing. I'd vote to leave it as is, just to make this case explicit.

PR #270, opened 2017-04-17
disable widget links (fixes #258)

### LINKED ISSUE #258: 404 when clicking on widget feature-by-stories

ehborisov:


plugins/behaviors-plugin/src/dist/static/index.js
```diff
@@ -13,7 +13,7 @@ allure.api.addTab('behaviors', {
     })
 });
 
-allure.api.addWidget('behaviors', allure.components.WidgetStatusView.extend({
+allure.api.addWidget('behaviors', Object.assign(allure.components.WidgetStatusView, {
```
baev: `Object.assign` will not work in IE. The problem is that `showLinks = true` in base class applied first. So you need to change `WidgetStatusView` like this:

```js
export default class WidgetStatusView extends View {
    template = template;

    serializeData() {
        const showLinks = typeof this.showLinks !== 'undefined' ? this.showLinks : true
        return Object.assign(super.serializeData(), {
            rowTag: showLinks ? 'a' : 'span',
            title: this.title,
            showAllText: this.showAllText,
            baseUrl: this.baseUrl
        });
    }
}
```

PR #295, opened 2017-04-28
screendiff improvements (fixes #294)

### LINKED ISSUE #294: Screen diff plugin brake test case layout when attachments missing

eroshenkoam:


just-boris:
+1. Your example pictures are Russian, it is not okay for documentation.

I would suggest to take screenshots of something more well-known, like https://www.google.de/#newwindow=1&q=unicorns

Issue #303, opened 2017-04-30
Change default categories name


eroshenkoam:
Seems like we can use "Defects tab" as the sample. 

Unknown failure -> Product defect
Unknown error -> Test defect

PR #304, opened 2017-04-30
Rename default categories

eroshenkoam:


eroshenkoam:
Possible we need to rename it to the defects plugin.

allure-generator/src/main/java/io/qameta/allure/category/CategoriesPlugin.java
```diff
@@ -115,6 +115,10 @@ public static boolean matches(final TestResult result, final Category category)
         return matchesStatus && matchesMessage && matchesTrace;
     }
 
+    private static boolean notPassed(TestResult result) {
```
baev: unused method

baev:
fixes #303

PR #305, opened 2017-05-01
Treeview tests

sseliverstov:


allure-generator/src/test/javascript/bootstrap.js
```diff
@@ -1,3 +1,11 @@
+import {Behaviors} from 'backbone.marionette';
+
+window.Marionette = require('backbone.marionette');
```
just-boris: Is that required? Seems like it should not.

sseliverstov: it is your magic =) without it I had exception: " TypeError: BehaviorClass is not a constructor
"

just-boris: Ok, we need to update one more line. Here is the entire code:

```
import {Behaviors, Behavior} from 'backbone.marionette';

Behaviors.behaviorsLookup  = function(options, key) {
 return { [key]: Behavior };
}
```

sseliverstov:
Tnx @just-boris, I fix it. @baev, merge?

just-boris:
Jenkins is failing

allure-generator/src/test/javascript/components/tree.spec.js
```diff
@@ -74,17 +73,12 @@ describe('Tree', function () {
...
-        this.el = renderView(data);
-        expect(this.el.paneTitle()).toMatch(tabName);
-    });
 
     describe('with empty data', function() {
 
         it('should render with empty data', function () {
             const el = renderView([]);
-            expect(el.tree().text().trim()).toMatch('There are no items');
+            expect(el.tree().text().trim()).toMatch('');
```
just-boris: it's better to assert that there are no nodes inside. Like so:

```js
expect(page.nodes()).toHaveLength(0)
```

sseliverstov:
Yep, sorry for the spam =(

sseliverstov:
@just-boris I fix it =)

just-boris:
happy to see that.
let's go forward!

baev:
thanks @sseliverstov @just-boris

PR #339, opened 2017-05-15
Add german translation

nscuro:
This addresses issue #329


pazone:
Vielen Dank! :)

baev:
@nscuro thanks!

PR #356, opened 2017-05-25
Update web deps

baev:


allure-generator/src/test/javascript/components/tree.spec.js
```diff
@@ -5,7 +5,7 @@ import settings from 'util/settings';
 import {values} from 'util/statuses';
 
 
-describe('Tree', function () {
+describe('Tree', () => {
```
just-boris: Please, keep all top-level describes as functions

allure-generator/src/test/javascript/components/tree.spec.js
```diff
@@ -116,12 +116,12 @@ describe('Tree', function () {
         ]
     });
 
-    describe('empty data', function () {
-        beforeEach(function () {
+    describe('empty data', function() {
```
just-boris: .... and make all nested as arrow.

So we will get one `describe('...', function() {})` per file.

PR #359, opened 2017-05-27
Fix attachment row hover (fixes #352)

### LINKED ISSUE #352: Wrong attachment hover
screen shot 2017-05-22 at 19 00 34

baev:


allure-generator/src/main/javascript/blocks/attachment-row/styles.css
```diff
@@ -1,7 +1,7 @@
 @import "../../variables.css";
 
 .attachment-row {
-  width: 90%;
+  width: 100%;
```
just-boris: Have you asked yourself a question why was it set to 90% in the first time?

baev: Seems like outdated stuff

just-boris: If you just remove this line, it will also work

PR #414, opened 2017-06-15
Add retry messages and fix flaky (fixes #413)

### LINKED ISSUE #413: Test with 100% passed rate displayed as flaky
#### I'm submitting a ... 
  - [x] bug report
  - [ ] feature request

My naive test runner doesn't do analysis and just restarts all tests on every run. As a result, all my tests are displayed as flaky even that one, which have always passed
screen shot 2017-06-15 at 6 21 30 pm

I expect to  have flaky tag only for those, which have passed rate around 50%

- Allure version:        2.1.1
- Test framework:       mocha
- Allure adaptor:        mocha-allure-reporter@1.3.2
- Generate report using: allure-commandline@2.1.1 from NPM

baev:
screen shot 2017-06-15 at 20 09 32

### DIFF
--- a/allure-generator/src/main/java/io/qameta/allure/retry/RetryPlugin.java
+++ b/allure-generator/src/main/java/io/qameta/allure/retry/RetryPlugin.java
@@ -3,6 +3,7 @@ package io.qameta.allure.retry;
 import io.qameta.allure.Aggregator;
 import io.qameta.allure.core.Configuration;
 import io.qameta.allure.core.LaunchResults;
+import io.qameta.allure.entity.Status;
 import io.qameta.allure.entity.TestResult;
 import io.qameta.allure.entity.Time;
 
@@ -15,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


allure-generator/src/main/java/io/qameta/allure/retry/RetryPlugin.java
```diff
@@ -47,26 +49,26 @@ public void aggregate(final Configuration configuration,
...
-        return retried -> {
-            retried.setHidden(true);
-            retried.getStatusDetailsSafe().setFlaky(true);
-            retries.add(new RetryItem()
-                    .withUid(retried.getUid())
-                    .withStatus(retried.getStatus())
-                    .withTime(retried.getTime())
-            );
-            latest.getStatusDetailsSafe().setFlaky(true);
+            latest.getStatusDetailsSafe().setFlaky(!statuses.isEmpty());
```
just-boris: Currently if a test fails 100% times, it will be displayed flacky. 
It is also wrong, I think

baev: 👎 

BTW flacky -> flaky

allure-generator/src/main/javascript/plugins/testcase-retry/RetryView.hbs
```diff
@@ -3,7 +3,8 @@
     {{#each retries}}
         <div class="retry-row" data-uid="{{uid}}">
             <span class="label label_status_{{status}}">{{status}}</span>
-            {{date time.start}} at {{time time.start}}
+            <span>{{date time.start}} at {{time time.start}}</span>
+            <div class="retry__message">{{statusDetails}}</div>
```
just-boris: if you will use `<p>`, then you can avoid extra styles, because `<p>` provides margins by default

PR #417, opened 2017-06-19
Show tooltip with step's start and finish times

leonsabr:
#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] ~~Provide unit tests~~


just-boris:
Very much related to https://github.com/allure-framework/allure2/pull/416

No rush, guys, seems like we have a common problem with timings. So let's find a common solution!

leonsabr:
> seems like we have a common problem with timings

What is the problem exactly?

just-boris:
We are displaying only duration, start and stop time is displayed in tooltip and not copyable.

Also, all time-related UI should be consistent, if we are going to implement copy-support for test duration, it has to be done here as well.

But anyway, this change is LGTM, let's keep this talk for #416

just-boris:
screen shot 2017-06-20 at 12 05 17 pm

@baev do you like the fact, that all tooltips are now declined to the left?

leonsabr:
So, there are 2 `time` elements on Test case view:
1) Test duration
2) Step duration

Consistent things:
1) All durations show milliseconds
2) MouseOver shows tooltip with text like `16.06.2017 17:01:01 - 17:02:04`

Inconsistent things:
1) Click on step's duration does not copy anything to clipboard
2) Tooltip for test's duration does not show milliseconds

I can fix first inconsistent thing, but I am not sure about the second one. Milliseconds may be critical for steps, but are not so critical for entire test.

leonsabr:
> all tooltips are now declined to the left

Current `bottom` behaviour makes some tooltips useless (you can't see the entire tooltip):

screen shot 2017-06-20 at 13 21 39

I have no idea how to make some View and it ChildView handle tooltips differently.

baev:
@just-boris I'm fine with new tooltip position.

But personally I don't like milliseconds in timings - it make it more unclear

@eroshenkoam thoughts?

baev:
@leonsabr thanks!

PR #426, opened 2017-06-26
Allow empty results directory (fixes #312)

### LINKED ISSUE #312: Allow empty allure results
Use-case:
I've got template jenkins task with allure report post-build action.

In some cases there is no tests and rest run produces no output in alluredir. In that case allure commandline fails and raises an Exception, and whole jenkins task is marked as failed.

Unfortunately it's hardly possible to turn off post-build action for some input parameters in jenkins, so I need allure-commandline not to fail in case of empty results.

baev:


just-boris:
Actually, I used to look at this message.
It usually said me, that there were problem with adapter.
For example, [integration tests in allure-mocha](https://github.com/allure-framework/allure-mocha/blob/master/test/e2e/main.spec.js#L24) will pass if adapter has crashed and did not produce any result.

just-boris:
What is the reason on change it? It is really breaking, if you did not know.

baev:
that change is required for `allure-gradle`, `allure-maven` and other integrations in case of multi-module projects. I mean if you have a module that have no tests yet, we can not generate a report. 

So for your case you will see an empty report, and that is a sign of some configuration problems.

just-boris:
Will you just skip the generation process at all, or generate a report with all zeroes?

baev:
we will generate empty report in case of no results found. Before this change you will get empty report anyway if your results directory exists but have no test results supported by Allure

just-boris:
Ok, I will follow up in allure-mocha project

PR #430, opened 2017-06-26
Fix description layout (fixes #419)

### LINKED ISSUE #419: Description layout is broken
#### I'm submitting a ... 
  - [x] bug report
  - [ ] feature request

screen shot 2017-06-20 at 16 49 54

#### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem
Just `./gradlew dev` and open demo report.

Less empty space between lines.

Common sense.

- Allure version:        `master`

baev:


just-boris:
How should this work for 
multi
line 
text

?

baev:
the same way it works for Github comments

just-boris:
I mean, what will happen for plain text? 
Or do you process via markdown by default?

PR #457, opened 2017-07-04
Tree refactoring, plugin api changes

baev:
- [x] Sign CLA
- [ ] Add unit tests


allure-generator/src/main/javascript/components/widgets-grid/WidgetsGridView.js
```diff
@@ -71,7 +71,8 @@ class WidgetsGridView extends View {
 
         this.addRegion(name, {el: el.find('.widget__body')});
        /// fetchAndShow(this, name, this.model, new Widget({model: this.model}))
-        this.getRegion(name).show(new Widget({model: this.model}));
+        const model = this.model;
+        this.getRegion(name).show(new Widget({model, name}));
```
just-boris: would be better to prepare the correct model here

```
new Widget({model: this.model.getWidgetData(name)})
```

instead of doing this inside the `Widget` constructor

allure-generator/src/main/javascript/data/tree/TreeCollection.js
```diff
@@ -69,4 +74,54 @@ export default class TreeCollection extends Collection {
...
         }
     }
+
+    calculateStatistic(items) {
+        const statistic = {};
+        values.forEach(value => {
+            statistic[value] = 0;
+        });
+        items.forEach(item => {
+            if (item.children) {
```
just-boris: you can also add  a check for `item.statistic` and don't do it on every `forEach` loop

allure-generator/src/main/javascript/data/tree/TreeCollection.js
```diff
@@ -69,4 +74,54 @@ export default class TreeCollection extends Collection {
...
+                time.sumDuration = time.sumDuration + item.time.duration;
+            }
+            TreeCollection.updateTime(time, item.time, 'start', Math.min);
+            TreeCollection.updateTime(time, item.time, 'stop', Math.max);
+            time.duration = time.stop - time.start;
+        });
+        return time;
+    }
+
+    static updateTime(timeA, timeB, field, operation) {
```
just-boris: This is not Java you don't need to use static methods, you can just create a standalone function within this module

allure-generator/src/main/javascript/data/tree/comparator.js
```diff
@@ -18,34 +24,34 @@ function byNodeStatus(a, b) {
...
-    if ('status' in a && 'statistic' in b) {
-        return () => 1;
-    } else if ('statistic' in a && 'status' in b) {
-        return () => -1;
-    } else if ('statistic' in a && 'statistic' in b) {
-        return groupCmp(a, b);
-    } else if ('status' in a && 'status' in b) {
-        return nodeCmp(a, b);
+function compare(a, b, nodeCmp, groupCmp, direction) {
+    if (a.hasOwnProperty('children') && !b.hasOwnProperty('children')) {
```
just-boris: why do you do this instead of just `if (a.children && !b.children) {`?

just-boris:
Also tested manually, key navigation was not broken 👍

allure-generator/src/main/javascript/components/widgets-grid/WidgetsGridView.js
baev: the problem is that graphs are widgets as well

just-boris: You mean this: https://github.com/allure-framework/allure2/blob/master/allure-generator/src/main/javascript/plugins/widget-status/StatusWidgetView.js
So, what is the problem?

baev: I mean that `graphs` tab uses `WidgetsGridView` but model is different (plain list of results)

PR #470, opened 2017-07-11
Add copy to clipboard behavior

baev:
- [x] Sign Allure CLA
- [x] ~Add unit tests~

screen shot 2017-07-11 at 20 37 29
screen shot 2017-07-11 at 20 37 34


allure-generator/src/main/javascript/components/testresult/TestResultView.js
```diff
@@ -15,6 +14,7 @@ const SEVERITY_ICONS = {
 };
 
 @behavior('TooltipBehavior', {position: 'left'})
+@behavior('ClipboardBehavior', {position: 'left'})
```
just-boris: How about built this option into behavior itself? It is not about copying and will never be changed

PR #476, opened 2017-07-15
use SASS instead of precss

just-boris:
There is no future in precss, also it makes use some problems in development, like bad support in Idea.

Sass would be better there.


allure-generator/build.gradle
```diff
@@ -8,7 +8,7 @@ apply plugin: 'maven'
 apply plugin: 'com.moowork.node'
 
 node {
-    version = '7.10.0'
+    version = '8.1.3'
```
just-boris: Not necessary for this particular feature, but good to have

allure-generator/webpack.config.js
```diff
@@ -35,17 +33,26 @@ module.exports = (env) => {
...
                         fallback: 'style-loader',
                         use: [
                             'css-loader',
-                            postcssLoader
+                            postcssLoader,
+                            'sass-loader'
                         ]
                     })
             }, {
+                test: /\.css$/,
```
just-boris: we still support plain css files, that are imported from libraries (e.g. font-awesome)

allure-generator/webpack.config.js
```diff
@@ -7,8 +7,6 @@ const postcssLoader = {
     loader: 'postcss-loader',
     options: {
         plugins: [
-            require('postcss-import'),
-            require('precss')({'import': {disable: true}}),
             require('autoprefixer')
```
just-boris: PostCSS is still used, as we want to have autoprefixer. (To have less pain supporting IE and Safari)

baev:
@just-boris thanks!

PR #321, opened 2017-05-07
add Chinese translation

wuhuizuo:
- add `translations/zh.json`
- register `zh.json` in util/translation.js

PR #487, opened 2017-07-21
Show attachments in modal view

sseliverstov:
[//]: # (
. Thank you so much for sending us a pull request! 
.
. Make sure you have a clear name for your pull request. 
. The name should start with a capital letter and no dot is required in the end of the sentence.
. To link the request with isses use the following notation: (fixes #123, fixes #321\)
.
. An example of good pull request names:
. - Add Russian translation (fixes #123\)
. - Add an ability to disable default plugins
. - Support emoji in test descriptions
)

### Context
[//]: # (
Describe the problem or feature in addition to a link to the issues
)

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/allure-generator/src/main/javascript/layouts/tree/TreeLayout.js
+++ b/allure-generator/src/main/javascript/layouts/tree/TreeLayout.js
@@ -2,6 +2,7 @@ import AppLayout from '../application/AppLayout';
 import TreeCollection from '../../data/tree/TreeCollection';
 import {Model} from 'backbone';
 import TestResultTreeView from '../../components/testresult-tree/TestResultTreeView';
+import router from '../../router';
 
 export default class TreeLayout extends AppLayout {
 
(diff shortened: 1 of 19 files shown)


allure-generator/src/main/javascript/layouts/tree/TreeLayout.js
```diff
@@ -28,5 +29,13 @@ export default class TreeLayout extends AppLayout {
     onRouteUpdate(testGroup, testResult, testResultTab) {
         this.routeState.set('treeNode', {testGroup, testResult});
         this.routeState.set('testResultTab', testResultTab);
+
+        const attachment = router.getUrlParams().attachment;
+        if (attachment) {
+            this.routeState.set('attachment', attachment);
+        } else {
+            this.routeState.unset('attachment', {silent: true});
```
just-boris: Silent is usually an anti-pattern. Can you not to use this option?

just-boris: You can add a check into `highlightSelectedAttachment` method instead.

just-boris:
Visual review

image

image

I don't like the current inline display. My suggestion

image

just-boris:
And my constant reminder about tab content: Steps should be on the first tab. I think this change should be done the first, before all other improvements with inline attachments.

sseliverstov:
@just-boris 
I am not sure that it is finally variant of layout for steps attachments. Your suggestion is interesting to, but in this iteration i save curent view.

just-boris:
* We have never use this blue color for borders. It just looks like 💩 . I tried to save it as much as I could.
* Modal still doesn't close if you will click on the place bellow the modal

sseliverstov:
@just-boris how about this?

sseliverstov:
@just-boris i am done

PR #600, opened 2017-09-20
Added search among testcases in html report

cakeinpanic:
We are using allure at work and suffer from lack of routing or at least search among test cases.
When we want to show someone specific allure report, we need to specify testcase and its parent, which is not very convenient

So i've added simple filtering on frontend, look how it works:
preview

I also need to add some unit tests, but as soon as I decided not to store`searchQuery` at `settings`(cause it shouldn't be remembered on page refresh), I couldn't write units fast.
Maybe you can give me some advice about it?

Looking forward for merge to start using it)


allure-generator/src/main/javascript/blocks/pane/styles.scss
```diff
@@ -12,6 +12,10 @@
 
   &__title {
     padding: 0 15px 10px;
+    width:100%;
+    display: table;
```
just-boris: Could we use `display: flex;` here? It would simplify the underlying code.

allure-generator/src/main/javascript/components/node-search/NodeSearchView.js
```diff
@@ -0,0 +1,27 @@
...
+import {View} from 'backbone.marionette';
+
+@className('search')
+class NodeSearchView extends View {
+    template = template;
+
+    initialize({onSearch, searchQuery}) {
+        this.onSearch = onSearch;
+
+        this.searchQuery = searchQuery;
```
just-boris: You could pass a model here instead of `searchQuery/onSearch`.

@baev do you like it?

allure-generator/src/main/javascript/components/tree/TreeView.js
```diff
@@ -50,25 +40,16 @@ class TreeView extends View {
...
-            settings: this.settings
-        }));
-        this.showChildView('filter', new StatusToggleView({
-            settings: this.settings,
-            statistic: this.collection.statistic
-        }));
+        if (this.searchQuery) {
+            this.$('.node__title').each((i, node) => {
+                this.$(node).parent().addClass('node__expanded');
+            });
```
just-boris: Something weird is happening here. 
@baev do we have expanded state stored in model?

allure-generator/src/main/javascript/blocks/pane/styles.scss
cakeinpanic: of course! i couldn't find information about browser support for report, so did it in oldschool way)

allure-generator/src/main/javascript/components/tree/TreeView.js
cakeinpanic: yes, when user clicks on a node, its state is stored, but i decided that better way is to force opening all nodes when search is in progress

cakeinpanic: when search is finished, i restore previous state.
I already see bug here – if user toggles nodes by click while he is in search mode, this state is saved and then will be restored incorrectly.

allure-generator/src/main/javascript/components/node-search/NodeSearchView.js
cakeinpanic: by the way,  searchQuery is not needed to be passed, it left from the time I tried to re-rended this componentn on each state change

allure-generator/src/main/javascript/components/node-search/NodeSearchView.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    onRender() {
+        this.$('input').val(this.searchQuery);
```
cakeinpanic: this should be gone, useless now

cakeinpanic:
Refactored using `settings` model, added unit test
We are really waiting for this feature, I'm interested in merging it)

@just-boris @baev @sseliverstov

allure-generator/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack --optimize-minimize",
     "start": "webpack-dev-server --inline --hot --env.development",
     "pretest": "eslint --ignore-path=.gitignore .",
-    "test": "jest",
+    "test": "jest --colors",
```
just-boris: It breaks console output in Jenkins logs.
It is better not to commit such changes.

allure-generator/src/main/javascript/components/node-search/NodeSearchView.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @on('input input')
+    onChangeSorting(e) {
+        this.settings.setSearchQuery(e.target.value);
+    }
+
+    close() {
+        this.settings.setSearchQuery('');
+    }
```
just-boris: Is this used?

allure-generator/src/main/javascript/components/tree-view-container/TreeViewContainer.js
```diff
@@ -0,0 +1,74 @@
...
+            state: this.state,
+            routeState: this.routeState,
+            tabName: this.tabName,
+            baseUrl: this.baseUrl,
+            settings: this.settings,
+            collection: this.collection
+        }));
+
+        this.showChildView('search', new NodeSearchView({
+            settings: this.settings
```
just-boris: Instead of settings, the `state` modal should be used. Then you will not need to introduce an extra `saveTemporary` method.

allure-generator/package.json
cakeinpanic: oh wow, couldn't imagine that can break anythig. For local development is a musthave option, otherwise log is not readable(

there probably shuoul be added some gradlew flag or something like this, i'll research this

allure-generator/src/main/javascript/components/node-search/NodeSearchView.js
cakeinpanic: not now, added just in case

allure-generator/src/main/javascript/components/tree-view-container/TreeViewContainer.js
cakeinpanic: I agree that extending `LocalStorageModel` is not the best way, but on the other side we store all settings(and searchQuery and filtering options are the same level settings) in one place. 
But i'll do as you say.

just-boris: We are doing different separation of concerns here:
* settings is for all data, that should be saved as user preference
* state is a current state of UI (opened tree nodes, selected testcase, now is also search). It will not be saved in LocalStorage, but can be reflected in the URL, for example

allure-generator/package.json
just-boris: As Gradle is involved here, I have nothing to say. Usually I ran tests via `npm test` directly out of `allure-generator` folder.

cakeinpanic:
added and fixed display bug

cakeinpanic:
found and fixed display bug, which caused tree re-render on each node folding

sskorol:
looks cool 👍

cakeinpanic:
@baev looking forward for merge & release)

allure-generator/src/main/javascript/data/tree/filter.js
```diff
@@ -18,4 +17,24 @@ function byDuration(min, max) {
 }
 
 
-export {byStatuses, byDuration};
\ No newline at end of file
+function byText(text) {
```
baev: byName

cakeinpanic:
Okay, waiting!
Our testers literally cant work with allure reports because of this(

baev:
@cakeinpanic thanks!

PR #603, opened 2017-09-21
Propose new link type "image"

nesterov-n:
This PR is for discussion purposes only.

First of all, thanks for your great tool!
It can build realy awesome reports and it would be even greater if it has ability to attach screenshot located at remote server.

I found that link can be attached to each test case. And there are some custom link types (`issue` and `tms`) that are rendered in special way.

Can we add one more special link type (`image` for example) that would be rendered as clickable image preview?
Something like that: 
screen shot 2017-09-21 at 20 38 37

Or maybe you can suggest another way to attach screenshots by link?

It seems to be related with issue #577 

Thank you in advance


nesterov-n:
Close #611 has implementation of remote attachments

PR #610, opened 2017-09-27
Add URI support for attachments

wesmcouch:
### Context
Allure attachments were always linked to local files on the report server. With this change you can then add an allure attachment in your test that links to a HTTP/s address of the attachment. For instance this change will allow you to upload your attachments to s3 and attach the link to the report. Our report sizes have begun to grow upwards of 2+ gb and our distributed testing methodology requires us to aggregate all of these files. This solves this by simply allowing each test to upload their attachment to s3 and reduce the size of the report to a couple megabytes.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/allure-generator/src/main/javascript/helpers/attachment-uri.js
+++ b/allure-generator/src/main/javascript/helpers/attachment-uri.js
@@ -1,0 +1,11 @@
+import {SafeString} from 'handlebars/runtime';
+
+export default function attachmentUri(source) {
+  let sourceUri;
+  if (source.startsWith('http')) {
+    sourceUri = source;
+  } else {
+    sourceUri = 'data/attachments/' + source;
+  }
+  return new SafeString(`<a class="link" href="${sourceUri}" target="_blank" data-tooltip="Open attachment in new tab">`);
+}
--- a/allure-generator/src/main/javascript/blocks/attachment-row/attachment-row.hbs
+++ b/allure-generator/src/main/javascript/blocks/attachment-row/attachment-row.hbs
@@ -12,8 +12,7 @@
             {{/if}}
         </div>
         <div class="attachment-row__control attachment-row__link">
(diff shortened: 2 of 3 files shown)


wesmcouch:
Are unit tests needed for this change?

allure-generator/src/main/javascript/helpers/attachment-uri.js
```diff
@@ -0,0 +1,11 @@
+import {SafeString} from 'handlebars/runtime';
+
+export default function attachmentUri(source) {
+  let sourceUri;
+  if (source.startsWith('http')) {
```
just-boris: startsWith doesn't exist in Internet Explorer. Use `source.indexOf('http') === 0`

allure-generator/src/main/javascript/helpers/attachment-uri.js
```diff
@@ -0,0 +1,11 @@
...
+    sourceUri = source;
+  } else {
+    sourceUri = 'data/attachments/' + source;
+  }
+  return new SafeString(`<a class="link" href="${sourceUri}" target="_blank" data-tooltip="Open attachment in new tab">`);
```
just-boris: Also, the helper should be responsible only for URL generation, not the `<a>` part. Just return `sourceUri` out here and leave opening `<a>` at the place where it was before

just-boris:
And the last question: would it be better to put this logic into report generator (e.g. Java part)? Let's generate final `sourceUrl`  at generation stage.

@baev ?

wesmcouch:
I added some more stuff into the Java part, I am going to re-open this PR with more changes.

allure-generator/src/main/javascript/helpers/attachment-uri.js
wesmcouch: I will make these changes in the new PR

PR #611, opened 2017-09-27
Add URI support for attachments

wesmcouch:
### Context
Allure attachments were always linked to local files on the report server. With this change you can then add an allure attachment in your test that links to a HTTP/s address of the attachment. For instance this change will allow you to upload your attachments to s3 and attach the link to the report. Our report sizes have begun to grow upwards of 2+ gb and our distributed testing methodology requires us to aggregate all of these files. This solves this by simply allowing each test to upload their attachment to s3 and reduce the size of the report to a couple megabytes.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


allure-generator/src/main/java/io/qameta/allure/allure2/Allure2Plugin.java
```diff
@@ -148,6 +148,12 @@ private Attachment convert(final Path source,
                 found.setName(attachment.getName());
             }
             return found;
+        } else if (attachment.getSource().startsWith("http")) {
```
wesmcouch: I probably need a null check here

wesmcouch:
Here is an example use case, by using a custom AllureResultsWriter, we can pipe the attachments into S3 and reformat the TestResult to contain the location in s3.

```
public class ExampleS3ResultsWriter extends FileSystemResultsWriter {

    private static final String AWS_BUCKET_LOCATION = "https://s3.amazonaws.com/allure-test-results/";

    public ExampleS3ResultsWriter(Path outputDirectory) {
        super(outputDirectory);
    }

    @Override
    public void write(TestResult testResult) {
        testResult.getSteps().forEach(step -> step.getAttachments().forEach(this::formatS3Location));
        testResult.getAttachments().forEach(this::formatS3Location);
        super.write(testResult);
    }

    private void formatS3Location(Attachment attachment) {
        attachment.setSource(AWS_BUCKET_LOCATION + attachment.getSource());
    }
    
    @Override
    public void write(String source, InputStream attachment) {
        // Upload to s3 instead of writing to file system
    }
}
```

I would also like to add documentation into the docs for others that want to take this approach if this is implemented.

allure-generator/src/main/javascript/helpers/attachment-uri.js
```diff
@@ -0,0 +1,11 @@
+import {SafeString} from 'handlebars/runtime';
+
+export default function attachmentUri(source) {
+  let sourceUri;
+  if (source.indexOf('http') === 0) {
```
chuckwilbur: In two other places your test is `source.startsWith('http')`. Would it be better to be consistent here?

wesmcouch: Yeah definitely, I made those adjustments

sskorol:
@baev @chuckwilbur @wesmcouch what's the state of this PR?

wesmcouch:
@sskorol It is ready to go, I haven't gotten any review from anyone yet on it.

baev:
Sorry for delay. 

I like the feature in general, but don't like the implementation. Seems like we need to introduce remote attachment in generator to process it proper way (I mean it would be nice to have at least a marker for such attachments)

BTW at the moment I'm working on huge core refactoring (realtime reporting, environments and more :tada: ) , and going to include the remote attachments feature to it.

wesmcouch:
@baev Thanks for the update and the hard work! I will go ahead and close this PR for now. If you want to add me to the pull request you open for remote attachments I can help review.

PR #650, opened 2017-10-29
Add separate jsons for widgets (fixes #567)

### LINKED ISSUE #567: Use separate json for each widget
[//]: # (
. Note: for support questions, please use Stackoverflow or Gitter**. 
. This repository's issues are reserved for feature requests and bug reports.
.
. In case of any problems with Allure Jenkins plugin** please use the following repository 
. to create an issue: https://github.com/jenkinsci/allure-plugin/issues
.
. Make sure you have a clear name for your issue. The name should start with a capital 
. letter and no dot is required in the end of the sentence. An example of good issue names:
.
. - The report is broken in IE11
. - Add an ability to disable default plugins
. - Support emoji in test descriptions
)

#### I'm submitting a ... 
  - [ ] bug report
  - [x] feature request

Currently all the widgets should be stored in single json file (see `widgets.json`, `graph.json`). It would be nice to have separate json for every widget. In that case we can reuse widgets on different tabs. 

| Allure version | any |
| --- | --- |
| Test framework | any |
| Allure adaptor | any |
| Generate report using | any |

#### Other information 

Also this change is required for #566 

[//]: # (
. e.g. detailed explanation, stacktraces, related issues, suggestions 
. how to fix, links for us to have more context, eg. Stackoverflow, Gitter etc
)

vbragin:


allure-plugin-api/src/main/java/io/qameta/allure/CommonJsonAggregator.java
```diff
@@ -17,9 +15,17 @@
  */
 public abstract class CommonJsonAggregator implements Aggregator {
 
+    private final String location;
+
     private final String fileName;
 
     protected CommonJsonAggregator(final String fileName) {
+        this.location = "data";
+        this.fileName = fileName;
```
baev: this("data", fileName)

allure-generator/src/main/java/io/qameta/allure/summary/SummaryPlugin.java
```diff
@@ -57,4 +39,28 @@ private SummaryData getSummaryData(final List<LaunchResults> launches) {
                 });
         return data;
     }
+
+    private static class JsonAggregator extends CommonJsonAggregator {
```
baev: is it used?

allure-generator/src/main/java/io/qameta/allure/status/StatusPlugin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Plugin that generates data for Graph tab.
+ *
+ * @since 2.0
+ */
+public class StatusPlugin extends CommonJsonAggregator {
```
baev: StatusChartPlugin

PR #651, opened 2017-10-29
Add pending (xfailed) status to report

gestory:


eroshenkoam:
@gestory we removed 'pending' status in allure 2 by design :) 
Which test are you mark as 'pending'?

gestory:
Hello!

Maybe I misunderstood the purpose of the status in allure, but in pytest
adapter the status is used for expected failures.

Do you have plans for displaying expected failures in allure?

Sincerely,
Denis Khodyakov

sseliverstov:
For pytest it is would be something like this: https://github.com/allure-framework/allure-python/issues/156

gestory:
Great, thank you, don't know how I missed this one. :)

Anyway, if my solution doesn't fit in your plans, can you tell me what you
are expecting as a solution?

Sincerely,
Denis Khodyakov

sseliverstov:
We have special fields for expected and muted failures in model 2(https://github.com/allure-framework/allure2-model/blob/master/allure2-model-pojo/src/main/resources/xsd/allure2_model.xsd#L131), but they are not supported in current ui version.

kolszewska:
Hi, would there be any follow up on this?

PR #670, opened 2017-11-16
Fix XcTestPlugin attachments saving in Xcode 9

azakordonets:
Problem: in Xcode 9 attachments now saving with `jpg` extension. Before it was stored with `png`. 
In order to fix this problem, i have added a small part of the code that checks for both `jpg` and `png` files in `Attachments` folder. If `png` is present, then we don't look for `jpg`. Otherwise, we try to find `jpg` attachment and add it to `Step`. 

Another change was made in `DurationTrendItem.java` - I faced with `Nullpointer` exception when I was testing locally, so I made a check and if time `duration` is `null`, then I set duration as 0. 

And finally, i made a small adjustment to `Contribution guide` - I have added steps that were missing for running project successfully in IntelliJ Idea.


plugins/xctest-plugin/src/main/java/io/qameta/allure/xctest/XcTestPlugin.java
```diff
@@ -151,6 +147,23 @@ private void parseStep(final Path directory, final Object parent,
         lastFailedStep.map(Step::getStatusTrace).ifPresent(step::setStatusTrace);
     }
 
+    private void addAttachment(final Path directory,
+                               final ResultsVisitor visitor,
+                               final Map<String, Object> props,
+                               final Step step) {
+        String uuid = props.get("UUID").toString();
+        Path pngAttachmentPath = directory.resolve("Attachments").resolve(String.format("Screenshot_%s.png", uuid));
```
baev: resolve attachments directory at first to remove duplicate "Attachments" string.

Also you can search for all files by `Screenshot_%s.*` glob and add them as attachments

azakordonets: Searching by `Screenshot_%s.*` didn't work, so instead i just loop over a extenstions array. Works locally for me

baev: did you try DirectoryStream? 

```java
try (DirectoryStream<Path> paths = Files.newDirectoryStream(attachments, String.format("Screenshot_%s.*", uuid))) {
    for (Path attachmentPath : paths) {
        if (Files.exists(attachmentPath)) {
            ...
        }
    }
} catch (IOException e) {
    ...
}
```

or you can use Stream API:

```java
Stream.of("jpg", "png")
        .map(ext -> attachments.resolve(String.format("Screenshot_%s.%s", uuid, ext)))
        .filter(Files::exists)
        .map(visitor::visitAttachmentFile)
        .forEach(step.getAttachments()::add);
```

azakordonets: Agree, Stream API looks really nice. Applied it and checked locally - works like a charm :) Should i squash all commits into 1 ?

plugins/xctest-plugin/src/main/java/io/qameta/allure/xctest/XcTestPlugin.java
```diff
@@ -20,12 +20,8 @@
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.time.DateTimeException;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
baev: do not use wildcard imports

plugins/xctest-plugin/src/main/java/io/qameta/allure/xctest/XcTestPlugin.java
baev: > Should i squash all commits into 1 ?

as u wish

plugins/xctest-plugin/src/main/java/io/qameta/allure/xctest/XcTestPlugin.java
azakordonets: Fixed, sorry about that

baev:
@azakordonets thanks!

azakordonets:
Sorry , for taking so long with such a minor thing :) I hope it will get to the next release

PR #673, opened 2017-11-19
Add new translation strings for German

just-boris:
New version introduces couple strings that should be re-translated


baev:
@just-boris thanks!

PR #676, opened 2017-11-23
Add JUnit's <system-out> element value to status message

aptester:
### Context
[//]: # (
Describe the problem or feature in addition to a link to the issues
)
JUnit's &lt;system-out&gt; element values are being skipped while generating the Allure report. Gherkin step details (Give/When/Then) are usually written to JUnit XML report using the &lt;system-out&gt; element tag which makes having them in the Allure report handy. 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


baev:
Is the status message right place to add such info? Maybe we should add `system-out` as attachment to the report?

aptester:
Looking at JUnit's xml report format, the <system-out> is the closest representative to log  full/intermediate status of the test case. Since for passing JUnit tests, the status message is currently empty, I thought it can be used effectively to show 'system-out' data. If the status message was already showing data returned by JUnit, then I would have picked attachment as well as this information could be termed extra content. Let me know if you agree.

aptester:
@eroshenkoam : Here is the sample JUnit.xml with Gherkin steps details that you requested.
[test-report.txt](https://github.com/allure-framework/allure2/files/1622193/test-report.txt)

eroshenkoam:
Sorry for such delay :(
It's look pretty good with small output:

But, I think it will be worse with big output. 

Maybe we can convert in into steps?

aptester:
@eroshenkoam , I didn't understand what you meant by converting the output into steps. As per me the output being displayed is of the 'steps' of the cucumber scenario. The feature file is which generated the above output is 
------
Feature: My Shared Calculator

  Background:
    Given my shared calculator is running

  @shared
  Scenario: Should subtract two positive numbers
    When I shared subtract 3 and 1
    Then shared result should be equal to 2
------
The output will always match the number of steps in the scenario. Hence I'm not worried about the length of the output because that is how the author of the scenario wrote it.

baev:
I had some changes in readers, and to prevent you from merging hell I will implement the feature by yourself. The issue to track https://github.com/allure-framework/allure2/issues/744

BTW thanks for all the work

PR #721, opened 2018-01-11
fixes #598

### LINKED ISSUE #598: Flaky tooltip is shown for non-flaky tests
#### I'm submitting a ... 
  - [x] bug report
  - [ ] feature request

When I am hovering cursor next to my parameters list I see a tooltip:

screen shot 2017-09-18 at 5 37 41 pm

Invisible icon still triggers tooltip, however it should not.

No icon – no tooltip

| Allure version | 2.2.0 |
| --- | --- |
| Test framework | testng@6.8 |
| Allure adaptor | allure-testng@2.0-BETA11 |
| Generate report using | allure-maven@2.18 |

symonk:
fixes #598, tooltips are flickering very badly and graphs too through out tho, can we look into that soon? I am happy to help but Im just learning js/hbs/frontend stuff

#### Checklist
- [X] [Sign Allure CLA][cla]

### DIFF
--- a/Analytics.md
+++ b/Analytics.md
@@ -22,6 +22,8 @@ every event:
 - The Google Analytics anonymous IP setting is enabled i.e. `1` (https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#aip)
 - The count of test results e.g `21` (https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#cm_)
 - The count of plugins used e.g. `11` (https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#cm_)
+- The test framework used e.g. `TestNG` (https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#cm_)
+- The language used e.g. `Java` (https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#cm_)
 - The Allure report analytics hit type is event (https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#t)
 
 With the recorded information, it is not possible for us to match any particular real user with the anonymized user ID. 


baev:
screen shot 2018-01-14 at 16 09 55

symonk:
If my styling preferences aren't suitable or bad UI/UX please throw me them out, I was getting to grips on front end design and learning so I was trying some stuff out

allure-generator/src/main/javascript/blocks/label/styles.scss
```diff
@@ -11,5 +11,7 @@
 
   &__info {
     background: #3c58e4;
+    margin: 1px;
```
baev: that is bad place to add margin and change display format, because this block can be used in many different places. Also info label should have the same margin as other labels such as failed or broken.

allure-generator/src/main/javascript/components/attachment/styles.scss
```diff
@@ -44,6 +44,7 @@
     }
   }
   &__media-container {
+    border: solid;
```
baev: `border: 1px solid $border-accent-color`

allure-generator/src/main/javascript/components/modal/ModalView.hbs
```diff
@@ -2,9 +2,7 @@
     <div class="modal__window">
         <h2 class="{{b cls 'title'}}">
             <span>{{title}}</span>
-            <div class="pane__controls">
-                <span class="modal__close fa fa-close" data-tooltip="Close"></span>
-            </div>
+            <button class="modal__close fa fa-close" data-tooltip="Close"></button>
```
baev: what is the point of the change? I personally dont like the button borders in here

allure-generator/src/main/javascript/components/modal/ModalView.js
```diff
@@ -28,7 +28,7 @@ class ModalView extends View {
         e.stopPropagation();
     }
 
-    @on('click .modal__background, .modal__close')
+    @on('click .modal__close')
```
baev: as far as I know usually modals can be closed by click on background, isn't?

allure-generator/src/main/javascript/components/tree/styles.scss
```diff
@@ -18,10 +18,11 @@
     position: absolute;
   }
   &__flaky {
-    padding-left: 10px;
```
baev: keep the padding here, `&_shown` style should only change the visibility.

allure-generator/src/main/javascript/components/tree/styles.scss
```diff
@@ -18,10 +18,11 @@
...
-    opacity: 0;
+    visibility: hidden;
     &_shown {
+      padding-left: 10px;
       opacity: 1;
```
baev: no need to change opacity at this point

allure-generator/src/main/javascript/components/modal/ModalView.hbs
symonk: is it ok to bring the X on the same line as the attachment name atleast? or do you prefer having the pane__controls div underneath, i guess if we add some new buttons later it may be best?

allure-generator/src/main/javascript/components/modal/ModalView.js
symonk: yes will revert, occassionally I would click off the attachment and close it accidently, but its quite a rare scenario and original behaviour is probably better

allure-generator/src/main/javascript/components/modal/ModalView.hbs
baev: > or do you prefer having the pane__controls div underneath

nope, thats a bug. X should be on the same line, float to the right

baev:
I guess the main issue with attachment modal is that there is no space between content and background:
 
screen shot 2018-01-14 at 18 21 04

Click on blur part of background should close the modal, on white - nope

baev:
or even
screen shot 2018-01-14 at 18 26 29

allure-generator/src/main/javascript/blocks/label/styles.scss
symonk: ok thanks, I will investigate the best solution for this, any recommendations here?

allure-generator/src/main/javascript/components/modal/styles.scss
```diff
@@ -18,6 +18,9 @@
     top: 0;
     z-index: 1001;
   }
+  &__close {
```
baev: it is better to use flex instead:

```hbs
<h2 class="{{b cls 'header'}}">
    <span class="{{b cls 'title'}}">{{title}}</span>
    <span class="{{b cls 'close'}} fa fa-close" data-tooltip="Close"></span>
</h2>
```

```scss
.modal {
  &__header {
    display: flex;
  }
  &__title {
    flex: 1;
  }
}
```

baev: also it would be nice to add translation for tooltip at this point. Simply replace `Close` text with the following template

```hbs
{{t 'controls.close'}}
```

and then you can add translations to `translations/*.json` files (only english is enough):

```
//en.json
...
"controls": {
   ...
   "close": "Close"
}
...
```

allure-generator/src/main/javascript/blocks/label/styles.scss
```diff
@@ -11,5 +11,6 @@
 
   &__info {
     background: #3c58e4;
+    padding-bottom: 1px;
```
baev: you can't change anything but color for this class. We have base style `label` that contains all the styles, and color modifier such as `label__failed`, `label__broken`, `label__info` etc.

baev: you can try to add padding for all the labels but I guess that is not a good solution

PR #726, opened 2018-01-17
(fixes #575 ) Minimizing effect of issue with expanding of previous test result when user clicks on different item

### LINKED ISSUE #575: If user click on a test items, other previously opened test packages also got expanded.
#### I'm submitting a ... 
  - [x] bug report

If user click on a test items, other previously opened test packages also got expanded.

#### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem
1. Open a test item on package A
2. Collapse package A
3. Open test item on package B

Package A stay collapsed. Only package B got expanded.

| Allure version | 2.3.4 |
| ---- | ---- | 
| Test framework | testng@6.11 |
| Allure adaptor | allure-testng@2.0-BETA16 |
| Generate report using | allure-jenkins-plugin@2.22 |

eichhornchen27:
fixes #575 
If user click on a test items, other previously opened test packages also got expanded. 
Fix works if all child nodes of the previously opened item are collapsed. Otherwise previous item will be expanded as a result of https://github.com/allure-framework/allure2/blob/master/allure-generator/src/main/javascript/components/tree/TreeView.js#L115

#### Checklist
- [x] [Sign Allure CLA][cla]

### DIFF
--- a/allure-generator/src/main/javascript/components/tree/TreeView.js
+++ b/allure-generator/src/main/javascript/components/tree/TreeView.js
@@ -92,8 +92,10 @@ class TreeView extends View {
         if (node) {
             const el = this.findElement(node);
             el.toggleClass('node__title_active', active);
-            this.changeState(node.testResult);
-            this.changeState(node.testGroup);
+            el.toggleClass('node__expanded', active);
+            el.parent('.node').toggleClass('node__expanded', active);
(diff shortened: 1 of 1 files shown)


baev:
the issue will be fixed in next major version via our big ui refactoring. 

BTW thanks for all the work

Issue #395, opened 2017-06-12
Add an ability to copy test case start-end timings to clipboard


AlexanderPoleschuk:
[//]: # (
. Note: for support questions, please use Stackoverflow or Gitter**. 
. This repository's issues are reserved for feature requests and bug reports.
.
. In case of any problems with Allure Jenkins plugin** please use the following repository 
. to create an issue: https://github.com/jenkinsci/allure-plugin/issues
.
. Make sure you have a clear name for your issue. The name should start with a capital 
. letter and no dot is required in the end of the sentence. An example of good issue names:
.
. - The report is broken in IE11
. - Add an ability to disable default plugins
. - Support emoji in test descriptions
)

#### I'm submitting a ... 

  - [x] feature request

#### What is the current behavior?
Test case details view has only an ability to view test case start-end values as a tooltip

#### What is the expected behavior?
It'd be nice to have an ability to copy info(start-end) from the tooltip to clipboard

#### What is the motivation / use case for changing the behavior?
Usability reasons

#### Please tell us about your environment:
- Allure version:        2.1.1
- Test framework:        testng@6.10
- Allure adaptor:        allure-testng@2.0-BETA8

#### Other information 

![2017-06-12_16-29-50](https://user-images.githubusercontent.com/9446141/27036152-96794568-4f8c-11e7-81fc-8f53e9963084.png)

jk816vu:
@baev Hi i was looking on this issue but i see that its been discused in https://github.com/allure-framework/allure2/pull/733. Is it still open ?

baev:
yes as far as I know

jk816vu:
@baev so can i work on it? And how it should look? something like this ? 
![image](https://user-images.githubusercontent.com/37405331/143584482-e51765e5-cf43-4af8-b94b-9ee49ebed67b.png)

jk816vu:
And what should be in clipboard after pressing button: 
data-tooltip a.k.a. "10/25/2021 17:55:12-17:55:13" ?

jk816vu:
Hi @baev and @just-boris i just made it work like clickable span like baev said. Can you check it ?

just-boris:
I am wondering why is this important to make the time copyable. You can always select the text and press ctrl+c on any content. Test suite and case name make sense, because they are the most used. But what about time?

[...]

PR #733, opened 2018-01-28
adding indicator for stack trace and english translation implements #501

symonk:
Adding a simple indicator that the statusMessage can be expanded to the user with a new icon + some text.  I am unsure of translations for other language(s) or how we go about implementing those, google translations? so I have added an english translation for now, waiting for your advice there.

### DIFF
--- a/allure-generator/src/main/javascript/blocks/status-details/status-details.hbs
+++ b/allure-generator/src/main/javascript/blocks/status-details/status-details.hbs
@@ -2,9 +2,8 @@
     <div class="status-details__content">
         {{#if statusMessage}}
             <div class="{{b 'status-details' 'trace-toggle' status=status}} clickable" data-tooltip="{{t 'testResult.status.trace'}}">
-                <pre class="status-details__message"><code>{{statusMessage}}</code></pre>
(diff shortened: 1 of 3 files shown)


baev:
@symonk we already have some improvements in rft-web2 branch (such change gonna be included in next major release)

image 2018-01-28 16 53 51

symonk:
ah sorry about that, I wasen't too sure

baev:
> ah sorry about that, I wasen't too sure

not a problem, thank you for helping us!

symonk:
no problem :) Front end stuff is tough haha, is the duration copy of any use?

baev:
Linked issue: https://github.com/allure-framework/allure2/issues/395

I have no problem with it, but @just-boris does (as far as I know). Feel free to create separate pull request with the change and lets discuss

just-boris:
Yep, my concern is the same. Extra copy button pollutes the page for no reason.

Instead of exposing an extra button, we can just change the duration format, so it will become copyable. Simple.

baev:
@symonk thanks anyway!

Issue #769, opened 2018-03-29
Report generator fails with exception if zulu timestamp


akranga:
#### I'm submitting a ... 
  - [x] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

#### What is the current behavior?
JUnit Report generator fails with ParseException if it tries to process zulu timestamp: `timestamp="2018-03-28T15:04:55Z"`.

#### Steps to reprocue
1. Create this file
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="test_all.x" errors="0" failures="0" tests="29319784" hostname="tbd" time="2.03698" timestamp="2018-03-28T15:04:55Z">
    <system-out/>
    <system-err/>
  </testsuite>
</testsuites>
```
2. Try to generate report with CLI
3. See the error

#### What is the expected behavior?
JUnit report generator should detect Zulu time format and should use: `DateTimeFormatter.ISO_DATE_TIME`
#### What is the motivation / use case for changing the behavior?
Zulu (utc) timestamp is common widely used. Different of tools can (like Catch2 for C++) can export their test results into this format, which is well understood by Jenkins (for instance) however not respected by Allure.

#### Environment:
| Allure version | 2.6.0 |

akranga:
I can create a PR if needed

baev:
> I can create a PR if needed

sounds good to me

akranga:
Is it possible if somebody would take a look at the PR #789? Happy to d follow ups if needed

PR #789, opened 2018-05-07
Zulu timestamp implemetnation (via #769)

akranga:
### Context
Feature in details has been discussed in #769

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/plugins/junit-xml-plugin/src/main/java/io/qameta/allure/junitxml/JunitXmlPlugin.java
+++ b/plugins/junit-xml-plugin/src/main/java/io/qameta/allure/junitxml/JunitXmlPlugin.java
@@ -25,6 +25,8 @@ import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


plugins/junit-xml-plugin/src/main/java/io/qameta/allure/junitxml/JunitXmlPlugin.java
```diff
@@ -127,11 +130,21 @@ private void parseTestSuite(final XmlElement testSuiteElement, final Path parsed
...
+        return LocalDateTime.parse(timestamp, dtf);
+    }
+
     private Long getUnix(final String timestamp) {
         if (isNull(timestamp)) {
             return null;
         }
-        final LocalDateTime parsed = LocalDateTime.parse(timestamp);
+        final LocalDateTime parsed;
+        if (timestamp.matches(ZULU_TIME_FORMAT)) {
```
baev: maybe we should use try/catch instead of matching by regexp? Like iterate over supported formats and trying to parse until found one that matches

akranga: See no problem with try...catch block this makes logic more bullet proof and more readable.  My rational behind regex was an assumption that matcher should be slightly easier for JVM rather than exception.

I'll update the code with try catch and fallback to zulu format. Do you think it will make sense to check regex before parse in catch block?

baev: The solution I have in mind is to have a list of different data formats and iterate over formats trying to parse. If no format found log an exception and use null value. This way it would be easy to add support for different formats and even and an extension point to support custom user formats (that would be useful in case of custom framework that logs in extended junit xml format)

baev: > and even and an extension point to support custom user formats (that would be useful in case of custom framework that logs in extended junit xml format)

PS no need to implement this part right now

just-boris:
I tested this fix with my results of my [CasperJS](https://github.com/casperjs/casperjs) tests.

Apparently, Javascript emits timestamp with milliseconds: `2018-05-31T14:05:25.155Z`. 

@akranga could you please add support for ISO 8601 format too?

baev:
fixed via https://github.com/allure-framework/allure2/pull/815

BTW @akranga @just-boris thanks!

PR #796, opened 2018-06-01
Add default direction to the document, fix the spinner

just-boris:
If you open current version of report, there will be no spinner, while page is loading.

With this fix, the spinner comes back:

screen shot 2018-06-01 at 3 02 20 pm


baev:
@just-boris thanks!

PR #885, opened 2019-01-21
Fix #884 forward to authentication page when session is expired

### LINKED ISSUE #884: Link to main page (Allure in left upper corner) doesn't forward to authentication page if session is expired.
#### I'm submitting a ... 
  - [x] bug report

In my case I use Allure from Jenkins installed at http://host/path/ and secured with one of authorizations strategies, so that when session is expired, jenkins automatically offers authentication screen.

Currently the home page is linked as href="#", gets translated to http://host/path/allure/# 
In this case, browser (Chrome) doesn't forward automatically to relogin, showing empty page.

Redirection to auth page, when session is expried
If link is not ending with anchor (#) then browser redirects to the auth page

Workaround: is going to the url field, deleting the # and resubmitting. 
It takes time, and every user must be informed about this workaround.

| Allure version | 2.8.1 |
| Jenkins version | 2.150.1 |

#### Other information 

[//]: # (
PR with a fix is following.
)

vinogradoff:
[//]: # (

Fix the #884 by changing forwarding from http://host/path/allure-path/# to http://host/path/allure-path/

It has the same effect for  pages without authorisation and for pages with active session, but helps in case authorised session was expired.

#### Checklist
- [x] [Sign Allure CLA][cla]

### DIFF
--- a/allure-generator/src/main/javascript/components/side-nav/SideNavView.hbs
+++ b/allure-generator/src/main/javascript/components/side-nav/SideNavView.hbs
@@ -1,5 +1,5 @@
 <div class="{{b 'side-nav' 'head'}}">
-    <a href="#" class="{{b 'side-nav' 'brand'}}">
+    <a href="." class="{{b 'side-nav' 'brand'}}">
         <span class="{{b 'side-nav' 'brand-text'}}">Allure</span>
     </a>
 </div>


baev:
@vinogradoff thanks!

PR #888, opened 2019-01-27
Bump ui packages

baev:


allure-generator/package.json
```diff
@@ -77,19 +82,26 @@
...
+    "prettier:format": "prettier --write 'src/main/javascript/**/*.{ts,tsx,css,scss}'",
+    "start": "npm run webpack:dev",
+    "webpack": "node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js",
+    "webpack-dev-server": "node --max_old_space_size=4096 node_modules/webpack-dev-server/bin/webpack-dev-server.js",
+    "webpack:build": "npm run cleanup && npm run webpack:build:main",
+    "webpack:build:main": "npm run webpack -- --config webpack/webpack.dev.js --env.stats=normal",
+    "webpack:dev": "npm run webpack-dev-server -- --config webpack/webpack.dev.js --inline --host=0.0.0.0 --port=3000 --env.stats=minimal",
+    "webpack:dev-verbose": "npm run webpack-dev-server -- --config webpack/webpack.dev.js --inline --port=3000 --profile --progress --env.stats=normal",
+    "webpack:prod": "npm run cleanup && npm run webpack:prod:main",
+    "webpack:prod:main": "npm run webpack -- --config webpack/webpack.prod.js --profile"
```
just-boris: I guess, you do not need that many different webpack commands. How about leaving only the `build`, like it was before?

allure-generator/webpack/webpack.common.js
```diff
@@ -0,0 +1,56 @@
+const webpack = require("webpack");
+
+const utils = require("./utils.js");
+
+module.exports = options => ({
+  cache: options.env !== "production",
+  resolve: {
+    extensions: [".js", ".json"],
+    modules: ["node_modules"],
```
just-boris: This `modules` setting is a default, can be removed

allure-generator/webpack/webpack.common.js
```diff
@@ -0,0 +1,56 @@
...
+  },
+  stats: {
+    children: false,
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      "process.env": {
+        NODE_ENV: `'${options.env}'`,
+        DEBUG_INFO_ENABLED: options.env === "development",
+        SERVER_API_URL: `''`,
```
just-boris: Do you use this constants? If not, remove.

allure-generator/webpack/webpack.dev.js
```diff
@@ -0,0 +1,68 @@
...
+    entry: ["./src/main/javascript/index.js"],
+    output: {
+      path: utils.root("build/www"),
+      publicPath: "/",
+      filename: "app.js"
+    },
+    module: {
+      rules: [
+        {
+          test: /\.(sa|sc|c)ss$/,
```
just-boris: You have only `.scss` files, right?
Let's make it `/\.scss$/` for the simplicity.

allure-generator/webpack/webpack.dev.js
```diff
@@ -0,0 +1,68 @@
...
+      new SimpleProgressWebpackPlugin({
+        format: options.stats === "minimal" ? "compact" : "expanded"
+      }),
+      new FriendlyErrorsWebpackPlugin(),
+      new webpack.HotModuleReplacementPlugin(),
+      new writeFilePlugin(),
+      new webpack.WatchIgnorePlugin([utils.root("src/test")]),
+      new WebpackNotifierPlugin({
+        title: "Allure Report"
+      })
```
just-boris: All these plugins can be removed. Unnecessary stuff.

allure-generator/webpack/webpack.prod.js
```diff
@@ -0,0 +1,86 @@
...
+const sass = require("sass");
+
+const commonConfig = require("./webpack.common.js");
+
+const utils = require("./utils.js");
+
+const postcssLoader = {
+  loader: "postcss-loader",
+  options: {
+    plugins: [require("autoprefixer")({ browsers: ["last 10 versions"] })],
```
just-boris: You don't need to specify `browsers` here, they should be picked up from the `package.json`

allure-generator/webpack/webpack.prod.js
```diff
@@ -0,0 +1,86 @@
...
+  },
+  output: {
+    path: utils.root("build/www"),
+    publicPath: "/",
+    filename: "app.js",
+  },
+  module: {
+    rules: [
+      {
+        test: /\.(sa|sc|c)ss$/,
```
just-boris: Same about SASS

allure-generator/webpack/webpack.prod.js
```diff
@@ -0,0 +1,86 @@
...
+        },
+      }),
+      new OptimizeCSSAssetsPlugin({}),
+    ],
+  },
+  plugins: [
+    new MiniCssExtractPlugin({
+      filename: "styles.css",
+    }),
+    new webpack.LoaderOptionsPlugin({
```
just-boris: This plugin is not needed

allure-generator/webpack/webpack.prod.js
```diff
@@ -0,0 +1,86 @@
...
+      minimize: true,
+      debug: false,
+    }),
+    new WorkboxPlugin.GenerateSW({
+      clientsClaim: true,
+      skipWaiting: true,
+    }),
```
just-boris: this plugin is **very bad** for Allure. Every report would attempt to register its own service worker, so you will open a report and you will see a cached data from another report. Fun, isn't it?

PR #940, opened 2019-05-03
Fix flaky and newFailed icons aligment issue in Suites view

ovcharenko-di:
-  Fix flaky and newFailed icons aligment issue in Suites view by reordering `tree__strut` div

Allure_TreeStrut

The order is now consistent with the order in tree-group.hbs 


baev:
@ovcharenko-di thanks!

PR #992, opened 2019-08-19
Fix allure.api.translate for plugins

v03oki:
My first github PR, probably code is not perfect

I am creating allure plugin (and newbie to js). Documentation to allure2 states, that allure.api.translate should translate, but returns undefined. Debugging in browser shows, that value perfectly translated and thrown away in this function. Let's fix it!


v03oki:
Looks like I can't merge it (I see red circle with cross):
`Merging is blocked
The base branch restricts merging to authorized users. Learn more about protected branches.`
GitHub help doesn't help here
Can anyone else merge it?

baev:
@v03oki thanks!

PR #1019, opened 2019-10-29
Fix to apply sorting only to suites at Suites tab

uarlouski:
### Context
1) From a user perspective it seems a bit weird that when we apply sorting at Suites level for example by duration, nested tests are also sorted by duration, and then, to see the actual test executions sequence, we have to sort once again by order. I believe that some additional sorting strategy should be added for tests to make usage more convenient.
2) These changes make sortable only suites (high level nodes) on the Suites tab

#### Checklist
- [x] [Sign Allure CLA][cla]
- [-] Provide unit tests (no tests found for JS in allure-generator module)


valfirst:
@eroshenkoam @baev could you please review the changes?

Issue #906, opened 2019-03-06
History of test limited to 5


tedrigofra:
In the Allure report there are the tabs Overview, History and Retries.
When I click on History then I see only 5 previous test results, although Jenkins has many more builds with Allure reports.
Why is it only displaying 5 previous test results in the History tab? How can I see more in the History tab? Do I have to configure that somewhere?
I use the latest Allure versions and JUnit 5.4 and Maven and JDK 11.

baev:
Duplicate of #347

Issue #347, opened 2017-05-18
Make number of items in history and history thread configurable


rovner:
Right now history plugin limits 5 last launches and history trend widget limits 20 last launches. It would be great to have this numbers configurable.

eroshenkoam:
It will be done in 2.1.0 in the end of the weak

johnmondeen:
Has this been fixed?
How can the number of history-elements be configured?

derinbay:
Now it's 20, but still not configurable. Any plan to add this feature?

fescobar:
@derinbay I think visually it's not so useful to have more than 20 items.

derinbay:
@fescobar Maybe, but it would be better to let user choose that. I would prefer about 100 for example.

fescobar:
@derinbay with this docker container you can have more than 20 results stored
https://github.com/fescobar/allure-docker-service#keep-history-and-trends

derinbay:
@fescobar thank you very much, i'll try that

askfor:
> @derinbay I think visually it's not so useful to have more than 20 items.

Visually it's not, but some times it's useful to look for some information in old builds, especially for a JOB that runs very frequently.

baev:
BTW you always can navigate to older report and see another 20 items of history

askfor:
> BTW you always can navigate to older report and see another 20 items of history

Didn't know that, thanks.

PR #1059, opened 2020-02-19
Change plugin history limit to 20

patrickmartin0:
Since there has been an ongoing conversation about the history limit, I decided to change the plugin limit to 20. This would probably be enough for now until there is a proper way to configure it. 

Related Issues are: 
https://github.com/allure-framework/allure2/issues/347
https://github.com/allure-framework/allure2/issues/906

This is a small change and I can't see why there would be a limit to 5 results so I'd be really happy if this would be merged.


nicon89:
Wouldn't it make more sense to have it configurable by environment variable with default value set to 20?

PR #1088, opened 2020-05-10
Fixed display of attachment with uid consisting only numbers.

TemaMak:
Fixed display of attachment with uid consisting only numbers.

#### Checklist
- [x] [Sign Allure CLA][cla]


baev:
@TemaMak thanks!

Issue #925, opened 2019-04-04
The 'flaky' bomb icon is not received in the executed test list.


zzt-chun:
`import time
import random
def test_0():
    time.sleep(random.randint(1, 3))
    if random.randint(1, 5) != 1:
        raise Exception('Broken!')`

![image](https://user-images.githubusercontent.com/44250961/55550922-17adc900-570c-11e9-8a31-ca95c751cd5a.png)

python 3.7.0
allure-pytest 2.6.1
pytest 4.3.1

corinazaharia:
I'm having the same problem when I run the tests in Jenkins - how did you fix it?

riddhik84:
I am having same issue. Any solution for this?

techhemant:
Facing the same issue:
Using Allure allure-2.13.0

zzt-chun:
There is a very stupid solution. After generating the report, use a script to modify the file that records the case names and delete the parts we don't want

PR #871, opened 2018-12-18
Add flaky and new failed test cases filter

skhomuti:
[//]: # (
. Thank you so much for sending us a pull request! 
.
. Make sure you have a clear name for your pull request. 
. The name should start with a capital letter and no dot is required in the end of the sentence.
. To link the request with isses use the following notation: (fixes #123, fixes #321\)
.
. An example of good pull request names:
. - Add Russian translation (fixes #123\)
. - Add an ability to disable default plugins
. - Support emoji in test descriptions
)

### Context
[//]: # (
Describe the problem or feature in addition to a link to the issues
)
Issue #858
New failed - tests that failed in the current run, but were passed in the previous
```
statuses.size() > 1
                && statuses.get(0) == Status.FAILED
                && statuses.get(1) == Status.PASSED;
```
Flaky - tests where in the last five launches the tests passed were after failed.
Applies only to failed tests in the current run.
```
statuses.indexOf(Status.PASSED) < statuses.lastIndexOf(Status.FAILED)
                && statuses.indexOf(Status.PASSED) != -1;
```
#### Checklist
- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests

[cla]: https://cla-assistant.io/accept/allure-framework/allure2

PR #1135, opened 2020-09-14
Don't set flakiness to false when other plugins have already set it to true (fixes #910)

### LINKED ISSUE #910: Flaky tests not displayed in 2.10.0
**Describe the bug**
Flaky test "bomb" icon is no longer displayed in 2.10.0

**To Reproduce**
1. Run flaky test (test that fails on first attempt and passes on retry)
2. Generate report using allure 2.10.0

**Expected behavior**
Flaky tests would be marked with a "bomb" icon on the report

**Screenshots**
screen shot 2019-03-07 at 3 42 52 pm copy

**Environment:**

| Allure version | 2.10.0 |
| --- | --- |
| Test framework | cucumber 4.2.6 |
| Allure adaptor | allure-cucumber4-jvm 2.10.0 |
| Generate report using | allure-maven@2.10.0 |

**Additional context**
Flaky tests are properly displayed when report is generated using allure 2.9.0

wuillaum:
### Context

This PR does not address all of the wants that have been discussed in https://github.com/allure-framework/allure2/issues/910 but it does fix the original issue. Because https://github.com/allure-framework/allure2/pull/871 added the functionality for the history plugin to determine flakiness, it would blow away any other plugins that set flakiness for a test suite. This behavior is a defect in itself because it only happened as HistoryPlugin is listed later than RetryPlugin here https://github.com/allure-framework/allure2/blob/b64437f8b5bf5fa318286c9ae136a40d87baa9c2/allure-generator/src/main/java/io/qameta/allure/DummyReportGenerator.java#L81-L113

Given that the [current documentation says that retried tests are marked as flaky](https://docs.qameta.io/allure/#_retries), I believe that this PR is necessary to make the docs accurate again.

I tested by modifying one of the demos that had an existing test which retried, and set one of the results to `Passed`
image
image

Here are more issues that could be closed after this PR is merged
https://github.com/allure-framework/allure2/issues/925
https://github.com/allure-framework/allure-java/issues/452 (although I don't think this fixes the 'skipped' issue)
https://github.com/allure-framework/allure-java/issues/463
https://github.com/allure-framework/allure-java/issues/127 (this one is super old but there appears to have been interest this year)

On the same note, I believe a valid feature request would be to allow for plugins to give a `FlakyReason` string to a flaky suite, and display this list of reasons when hovering over the flaky bomb (whereas now it only says `Test is flaky`).

Edit: It would also seem that any custom plugins the users attempted to use would also have their flakiness blown away as custom plugins are loaded first in the array 
https://github.com/allure-framework/allure2/blob/b64437f8b5bf5fa318286c9ae136a40d87baa9c2/allure-generator/src/main/java/io/qameta/allure/DummyReportGenerator.java#L141-L148

 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/allure-plugin-api/src/main/java/io/qameta/allure/entity/TestResult.java
+++ b/allure-plugin-api/src/main/java/io/qameta/allure/entity/TestResult.java
@@ -71,6 +71,27 @@ public class TestResult implements Serializable, Nameable, Parameterizable, Stat
     protected boolean retry;
     protected final Map<String, Object> extra = new HashMap<>();
 
+
+    public TestResult resetFlaky() {
+        this.flaky = false;
+        return this;
+    }
+
+    /**
+     * Sets the TestResult's flaky status
+     * 
+     * Performs logical OR on existing flakiness so that other plugins that
+     * have previously set flakiness don't get their value overwritten 'falsely'
+     *
+     * Trying to deliberatey set flakiness to false? use {@link #resetFlaky()}
+     * 
+     * @param flakeStatus Flaky status
+     */
+    public TestResult setFlaky(final boolean flakeStatus) {
+        this.flaky = flakeStatus || this.flaky;
+        return this;
+    }
+
     @JsonProperty
     public String getSource() {
         return getUid() + ".json";


wuillaum:
@baev I'm not sure who to notify to look at this PR, it looks like you're the most active contributor to this repo.

lukaville (approved):

baev:
Thanks for the contribution! 

> Given that the current documentation says that retried tests are marked as flaky,

Having retries does't mean that test is flaky, isn't? Maybe better to introduce new marker specially for retries (🔄) ?

wuillaum:
It's up for interpretation. For me, a test that failed and retried from the perspective of a test-runner is a flaky test, as you should have internal retry mechanisms to catch/retry expected flaky behaviors in a test suite. 

Optimally you have zero failed tests, with safeguards built into the tests themselves. This way you're not relying on a test runner to retry to pass. 

I agree a retry marker would likely be better, but I figured fixing the code to reflect the originally intended behavior (based on the docs) would be best in the meantime. This PR also addresses the fact that no other plugin is allowed to determine flakyness (like if you wanted to have @flaky suites *always* show up in allure), which is definitely a bug.

baev:
> For me, a test that failed and retried from the perspective of a test-runner is a flaky test

The problem is that you may retry passed test for some reason. Or, for example, if you keep allure-results uncleared between runs (usually happens on local test runs) you'll get invalid flaky markers on your tests. So we ended up changing this to analysing history instead of retries (and docs are just outdated).

Plus, imagine user that uses `@ RepeatedTest ` annotation - then all such tests will become flaky no matter of what.

From my point of view flaky should be only set if test has different fail reasons when run on same environment. But this one is tricky to implement as you'll never know -- probably in previous run that was just another defect, that was fixed between runs and now your test found another one.

Currently we are working on different analytics features in our paid product (Allure EE), trying to build a reasonable metrics for test cases, and some of this metrics will be implemented in Allure report as well. 

Speaking of changes you provided -- I'm fine with having a possibility to override flaky mark using plugins, but I rather use separate marker for retries to keep current behaviour. At least we should let users decide what policy to use to mark tests as flaky.

cpjust:
Oh for gods sake!  The definition of a flaky test is a test that has **inconsistent results when you re-run it**.  If you run a test 5 times and it fails all 5 times, it's not flaky, it just failed and probably indicates either a bug in the test or a bug in the product you're testing.  If it passes all 5 times, it's not flaky.  But if it has a mix of pass and fail results, it's flaky -- plain and simple.

The isFlaky() function is only checking if the _first_ status is failed:
```
private boolean isFlaky(final List<HistoryItem> histories) {
    if (histories.size() > 1 && histories.get(0).status == Status.FAILED) {
```

It should check if _any_ status in the history is different than the others.  So something like this:

```
private boolean isFlaky(final List<HistoryItem> histories) {
    if (histories.size() > 1 && (histories.stream().filter(item -> item.status).distinct().count() > 1)) {
```

It should do this for both historical runs and re-runs in the same test run.

Issue #1240, opened 2021-04-27
"new broken" filter


redsun82:
**Is your feature request related to a problem? Please describe.**
The "new failure" filter is quite convenient, it would be nice to have exactly the same way to quickly spot the tests that just got into the "broken" status.

**Describe the solution you'd like**
A "new broken" quick filter just like the "flaky" and "new failed" ones that would list the tests that are broken, but were passing in the previous run.

**Describe alternatives you've considered**
A workaround to have something similar would be in a specific project to force broken tests to appear as failed (for example, a pytest fixture turning all exceptions into assertion errors), so that new broken would appear alongside the new failures. One would however then lose the distinction between the two states.

**Additional context**
Nothing more to add.

fludkov:
I completely agree with this feature request as I have a lot of faled and broken tests in my suite and I miss some new broken tests. Actually they mark as broken as selenium did not found some element on page (layout of page changed) and my tests did not throw AssertException on this but this is also I need to track. Please implement this filter

PR #1290, opened 2021-05-26
Implement "New Broken" filter (fixies #1240)

fludkov-valerii:
- I`ve not changed Flaky function to support newBroken feature but it would be useful
- Please note, that I have no idea how to translate "New broken" to kr and ja languages


fludkov:
updated commit b056364 with fixed line length in allure-generator/src/main/javascript/data/tree/filter.js file (line 35). Please run workflows again

baev:
I like that feature in general. What do you think, maybe it would be useful to not only highlight new broken tests but all the results with the status change? I mean:


from status | to status | comment
--- | --- | ----
failed or broken | passed | fixed tests
passed | failed | new failure
passed | broken | new broken
no result | any | new tests
any | no results | missed

fludkov:
> I like that feature in general. What do you think, maybe it would be useful to not only highlight new broken tests but all the results with the status change? I mean:
> 
> from status	to status	comment
> failed or broken	passed	fixed tests
> passed	failed	new failure
> passed	broken	new broken
> no result	any	new tests
> any	no results	missed

Yes, quick filtering by fixed tests would be very useful as managers usually asks if tests has passed after deploying fix to some environment. 
Besides after implementing new test I want to check how it passed in CI and which result it has. Filter for this would be very convenient. 

But I`m afraid I will not able to implement these new statutes as I'm a python programmer and have no much experience in Java :)

PR #1497, opened 2021-11-26
Clickable solves #395

jk816vu:
#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/allure-generator/src/main/javascript/plugins/testresult-duration/DurationView.hbs
+++ b/allure-generator/src/main/javascript/plugins/testresult-duration/DurationView.hbs
@@ -2,6 +2,6 @@
     <span data-tooltip="{{date time.start}} {{time time.start}}&nbsp;&ndash;&nbsp;{{time time.stop}}">
         {{t 'testResult.duration.name'}}:
         <span class="fa fa-clock-o"></span>
-        {{duration time.duration 2}}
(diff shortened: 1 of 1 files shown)


jk816vu:
i have no idea how to add label

jk816vu:
@baev Hi i also wanted to update tooltip for few seconds 
`this.setAttribute('data-tooltip', 'Text copied!')`
Problem is when i do it in web browser console, it updates without issue. 

If i do it inside `onClick` it does not update the view of the affected tooltip. When i later print out the `data-tooltip` value in console instead it prints out my new value. What i am trying to say is that value in data-tooltip does not match what is in console. What could be the issue here ? 
on website:
image
in console:
image

allure-generator/src/main/javascript/plugins/testresult-duration/DurationView.hbs
```diff
@@ -2,6 +2,8 @@
     <span data-tooltip="{{date time.start}} {{time time.start}}&nbsp;&ndash;&nbsp;{{time time.stop}}">
         {{t 'testResult.duration.name'}}:
         <span class="fa fa-clock-o"></span>
-        {{duration time.duration 2}}
+        <span
+        onclick="navigator.clipboard.writeText('{{date time.start}} {{time time.start}} - {{time time.stop}}');"
```
baev, line 6: please use `{{> ../../blocks/clipboard-copy/clipboard-copy value=...}}`  instead. See https://github.com/allure-framework/allure2/blob/8aa23ba7d84ccb4742f499bf4a79e2255f943dca/allure-generator/src/main/javascript/components/testresult/TestResultView.hbs#L3

baev (changes requested):

PR #1544, opened 2022-01-04
Add support for supplying input zip files next to directories

miensol:
With this change it is possible to run
`allure generate --report-dir ./out ./source-1.zip ./source-2.zip ./source-3`

### Context

We're building a POC that would bring together tests results from various CI jobs. 
We've attempted to run allure from AWS Lambda. Raw tests results are synced from S3 to an EFS attached to Lambda. The performance of generating the report is terrible when compared to local execution. It turned out that the majority of the time the lambda spends on checking files metadata. This is understandable as allure needs to list files and directories. 

In order to tackle this, we'd like to zip test results i.e. thousands of files, sync the archives to the EFS attached to the Lambda instead of raw test results. 

The way it was implemented was to run readers (except the zip results source plugin) again after opening the zip archive.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/allure-generator/src/main/java/io/qameta/allure/ConfigurationBuilder.java
+++ b/allure-generator/src/main/java/io/qameta/allure/ConfigurationBuilder.java
@@ -50,6 +50,7 @@ import io.qameta.allure.suites.SuitesPlugin;
 import io.qameta.allure.summary.SummaryPlugin;
 import io.qameta.allure.tags.TagsPlugin;
 import io.qameta.allure.timeline.TimelinePlugin;
+import io.qameta.allure.zip.ZipResultsSourcePlugin;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


noomorph:
The feature looks promising! However, I was wondering if it would be possible to add equal support for `*.tar` and `*.tar.gz` archives as well.

The motivation behind TAR archives is especially interesting for custom Allure runtime writers. If a writer appends files (like `9283749-result.json` and `2384978923-container.json`) straight to a tar archive instead of a directory, it can unload the filesystem from operations related to directory/file management, without even having to compress things. I believe this could significantly improve performance, especially on 5000-6000 test suites.

A1exKH (changes requested):
Code LGTM.

Please resolve conflicts for this PR.

CloneOfAlex (approved):

CloneOfAlex (approved):

felseje:
Hey, please, resolve conflicts for this PR.

baev:
Allure 3 supports report generation from dump archives

Issue #1638, opened 2022-04-27
Please add Info tag also right now we have only 5 tag (passed,failed,broken,unknown,skipped)


lalitdhami:
**Is your feature request related to a problem? Please describe.**
 when i am trying to show info log in my allure report i don't have any option to do that right now i am using unknown tag but it is not good to show info like that

**Describe the solution you'd like**
Please add Info tag

PR #1642, opened 2022-05-01
added Info tag as right now we have only 5 tag (passed,failed,broken,…

dheerajsingh544:
…unknown,skipped)

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

#https://github.com/allure-framework/allure2/issues/1638
Fixes #


lalitdhami:
i think this code will not work . only one enum change will not work we need to change js and other also . i have done that will raise that pr . please delete your pr

PR #1829, opened 2022-11-14
Improve xray plugin with jira request pagination (fix #1828)

### LINKED ISSUE #1828: Xray plugin throws error when Test execution contains more than 1000 test cases.
We use Jenkins Xray plugin to submit test results to the Xray. Allure submits results and only after that generates the report. 
Currently we faced the situation when the test execution we want to sumbit results to has more than 1000 test cases. Jira has max results limit within the api calls so when the Allure plugin tries to get all the test cases from Xray execution it receives an error from Jira 

image

And hence the report is not generated because the previous stage has failed.

To fix the issue I suggest to modify the getTestRunsForTestExecution() service method and include page query parameter as it described in Xray api docs

image

fedavar:
Resolve the issue with jira "Maximum results per request exceeded" for tests in test execution. fix #1828


fedavar:
Please excuse me for the redundant commits, I will create a new PR for the fix

PR #1833, opened 2022-11-17
Map JUnit XML test case `<properties>` to test result parameters

Holzhaus:
### Context
This makes a test case's `<properties><property name="foo" value="bar" /></properties>`  element in the Junit XML file show up as test case parameters in the report.

#### Checklist
- [X] [Sign Allure CLA][cla]
- [X] Provide unit tests


Holzhaus:
Ping. Any chance to get this merged? The failing check does not look like something I'm able to change (I cannot edit labels).

baev:
@Holzhaus, could you please clarify what test framework has such output?

Holzhaus:
For example:
- pytest - https://github.com/pytest-dev/pytest/blob/main/src/_pytest/junitxml.py#L150-L153
- jest - https://github.com/jest-community/jest-junit/issues/135

Also, [xunit-viewer](https://github.com/lukejpreston/xunit-viewer) (another tool for creating an HTML report from a JUnit XML report) supports displaying these properties, so I figured that Allure2 should be able to do this as well.

baev (approved):

baev:
@Holzhaus thanks!

PR #1145, opened 2020-10-18
screen-diff-plugin: support multiple screen diffs per test via custom attachment view

just-boris:
### Context

closes #613

Allow multiple screenshot differences via attachments. It should have the following format

```json
{
  "expected": "data:image/png;base64,....",
  "actual": "data:image/png;base64,....",
  "diff": "data:image/png;base64,...."
}
```

To support this, I extended plugin API allowing to register custom attachment views per mime type

```js
allure.api.addAttachmentViewer(mimeType, {
  View: CustomView,
  icon: 'fa  fa-exchange', // custom icon if you want
});
```

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests – where are all frontend tests? Can't find any

[cla]: https://cla-assistant.io/accept/allure-framework/allure2

PR #1869, opened 2022-12-21
Take nested attachments into account when doing the screenshots diff

nonatomiclabs:
### Context

The screen-diff plugin works great but it doesn't take into account attachments that are nested in an Allure step, or that are attached to the before-/after-test stages. This PR extends the `findImage` function of the plugin to take all the steps/stages into account.

I tested it locally, as can be seen on the following screenshot (black boxes hide private information, red lines highlighting the change):
allure-screenshot

#### Checklist
- [X] [Sign Allure CLA][cla]
- [ ] Provide unit tests: not sure how to add them, as I didn't find tests for this plugin

### DIFF
--- a/plugins/screen-diff-plugin/src/dist/static/index.js
+++ b/plugins/screen-diff-plugin/src/dist/static/index.js
@@ -12,8 +12,37 @@
     }
 
     function findImage(data, name) {
-        if (data.testStage && data.testStage.attachments) {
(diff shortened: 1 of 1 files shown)


baev:
Screen Diff plugin is deprecated. We suggest you use imagediff attachments instead. See https://github.com/allure-framework/allure2/pull/1145 for details

nonatomiclabs:
Thanks a lot, I'll have a look at imagediff!

PR #1927, opened 2023-02-27
Performance refine for convert jsons to website.

XenoAmess:
### Context

Hi.

I have a repo who have 3000+ testcases, and we try use allure2.

But it is slow, so we did some profiling and made such a refine patch.

Please have a look, thanks.

PS1. Before this patch the time cost to generate the website is around 30 min, and after it becomes 15 sec.

PS2. This patch is necessary to me as according to plan this repo's testcases would grow to at least 15000+ to make it a 95%+ coverage

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/allure-generator/src/main/java/io/qameta/allure/allure2/Allure2Plugin.java
+++ b/allure-generator/src/main/java/io/qameta/allure/allure2/Allure2Plugin.java
@@ -44,9 +44,14 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


XenoAmess:
add label improvement

XenoAmess:
@baev checkstyle and pmd fixed.
please review again, thanks.

baev:
@XenoAmess, thanks for the contribution. I'll need a bit more time to test & review it

baev:
@XenoAmess good catch. Please review this change https://github.com/allure-framework/allure2/pull/1982. Do you think it suits your needs? I tested it on my killer results set, and it is now much faster than before. 

The main difference with the change you provided is that I dropped support for recursive fixtures (it is never used and not supported by Allure TestOps as well) , so the code is now much simplier

PR #2445, opened 2024-03-25
Add Swedish translations

atg-pablo-widlert:
### Context
Add Swedish translations for Allure

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests (not applicable)

[cla]: https://cla-assistant.io/accept/allure-framework/allure2

PR #2476, opened 2024-04-04
Add possibility to specify report language via cli option

baev:
### Context

Now you can change the default report language using the `--lang` option:

```bash
$ allure generate --lang de path/to/allure-results
```

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/allure-commandline/src/main/java/io/qameta/allure/validator/LanguageValidator.java
+++ b/allure-commandline/src/main/java/io/qameta/allure/validator/LanguageValidator.java
@@ -1,0 +1,60 @@
+/*
+ *  Copyright 2016-2024 Qameta Software Inc
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package io.qameta.allure.validator;
+
+import com.beust.jcommander.IParameterValidator;
+import com.beust.jcommander.ParameterException;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author charlie (Dmitry Baev).
+ */
+public class LanguageValidator implements IParameterValidator {
+
+    private static final Set<String> SUPPORTED_LANGUAGES = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList(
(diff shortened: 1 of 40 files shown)


allure-commandline/src/main/java/io/qameta/allure/validator/LanguageValidator.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author charlie (Dmitry Baev).
+ */
+public class LanguageValidator implements IParameterValidator {
+
+    private static final Set<String> SUPPORTED_LANGUAGES = Collections.unmodifiableSet(
```
noomorph, line 31: Is there a chance to make it aware of `i18n/<lang>` availability?

allure-generator/src/main/javascript/utils/settingsFactory.js
```diff
@@ -35,7 +34,7 @@ const widgetGridPluginDefaults = {
 function getGlobalSettings() {
   const SettingsModel = LocalStorageModel.extend({
     defaults() {
-      return globalSettingsDefaults;
+      return { ...globalSettingsDefaults, language: document.documentElement.lang || "en" };
```
noomorph, line 37: I have a question.

Would you consider adding an option like `auto` for language?

This can work in a similar way to this:

https://github.com/sonic16x/interslavic/pull/222/files#diff-6c9b6c5e0990d5094cdffba8ac89a46c924ab7f2a69ba718d819c95118e3ee77

The idea is to fall back to English when there's no good match with `navigator.languages`. However, if there is a match, you can set the language to the user's most preferred one.

baev: As far as I remember, we used user-preferred language before, and the community asked us to use English by default instead. It may make sense after this PR is merged (so "auto" by default, by you can change it to whatever language you want by using the generator option ), although I don't have such a feature request in my backlog at the moment.

allure-commandline/src/main/java/io/qameta/allure/validator/LanguageValidator.java
baev: We can, but it will be tricky to implement. IMHO, it is not worth the effort

noomorph: Well, I just thought that the contributors adding new languages, like in https://github.com/allure-framework/allure2/pull/2445, may be forgetting to update this thing because it's not very obvious. In any case, this can be solved with documentation, indeed.

allure-generator/src/main/javascript/utils/settingsFactory.js
noomorph: > so "auto" by default, by you can change it to whatever language you want by using the generator option

I would even suggest keeping "en" as the default language, unless "auto" is specifically configured. 😄

Regardless, this feature is very useful. I have been wondering if it is possible to set a default language ever since I added Interslavic translations to the project. Good to see things moving. 👍

eroshenkoam (approved):

PR #2541, opened 2024-05-08
* Images are found not only in Test Stage but also in sub steps

fatih-sukran:
### Context
In order for the image comparison plugin to work, the expected, actual and diff images must be in the attachments under the Test Stage. With the development, it can also find expected, actual and diff suffixes under any Step.

Screenshot 2024-05-08 at 10 20 37

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/plugins/screen-diff-plugin/src/dist/static/index.js
+++ b/plugins/screen-diff-plugin/src/dist/static/index.js
@@ -19,7 +19,27 @@
             if (matchedImage) {
                 return 'data/attachments/' + matchedImage.source;
             }
+            return findImageInSteps(data.testStage.steps, name)
         }
+        return null;
+    }
+
+    function findImageInSteps(steps, name) {
+        for (var index = 0; index < steps.length; index++) {
+            var matchedImage = steps[index].attachments.filter(function (attachment) {
+                return attachment.name === name;
+            })[0];
+
+            if (matchedImage) {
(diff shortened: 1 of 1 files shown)


baev:
Thanks for your contribution!

The API you're using is deprecated. A new [Visual Comparisons](https://allurereport.org/docs/attachments/#visual-comparisons) feature allows you to have multiple diffs per test.

It works the same way, but instead of adding multiple attachments to the root level, you need to create a single JSON that will contain all the images (actual, expected and the diff) in base64:

```js
  // Read the three files and encode to base64
  const expected = fs.readFileSync("expected.png", "base64");
  const actual = fs.readFileSync("actual.png", "base64");
  const diff = fs.readFileSync("diff.png", "base64");

  // Wrap into a JSON, encode as string
  const content = JSON.stringify({
    expected: `data:image/png;base64,${expected}`,
    actual: `data:image/png;base64,${actual}`,
    diff: `data:image/png;base64,${diff}`,
  });

  // Attach to the test report
  allure.attachment("Screenshot diff", content, "application/vnd.allure.image.diff");
```

fatih-sukran:
Hi @baev, 
I didn't know how to use the new visual comparison feature. 
This does what I want to do. I no longer need PR. I'm closing PR. 
Thanks.

PR #2556, opened 2024-05-09
Fix RPM JRE dependency

KoLLlMapuK:
### Context
Rpm package was not installed due to an incorrect dependency (default-jre-headless). In rpm repo this dependency called "jre"


baev (approved):

baev:
@KoLLlMapuK thanks

Issue #889, opened 2019-01-30
Add timestamp of last report to prometheus metrics


koliyo:
**Is your feature request related to a problem? Please describe.**
Yes, I want to be able to detect if tests are not being executed, eg if the test runner is failing. If this happens do not get a report and thus  there are no failing tests. But I want to detect this case, by monitoring the timestamp of the last successful test 

**Describe the solution you'd like**
Store the timestamp of the last test report in the exported prometheus metrics

**Describe alternatives you've considered**

**Additional context**

PR #2626, opened 2024-07-03
Expose "start" and "stop" timestamps as metrics

pfyod:
### Context

In order to be able to track whether tests actually ran, there should be some sort of timestamp in Prometheus / influxdb metrics.

This PR exposes existing `start` and `stop` timestamps from `GroupTime` as metrics inside the `TimeMetric` class. I guess they were simply overlooked as all other properties of  `GroupTime`  are already exposed.

Three are no existing unit tests for the involved classes, so not sure where would one add them and what to test there.

This fixes an existing issue #889.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests


baev (approved):

baev:
@pfyod thanks!

PR #3335, opened 2026-05-10
Preserve testCaseId from on-disk schema as entity.TestResult#testId

grootstebozewolf:
### Context

The on-disk Allure 2 schema (defined in [allure-java's `io.qameta.allure.model.TestResult`](https://github.com/allure-framework/allure-java/blob/master/allure-model/src/main/java/io/qameta/allure/model/TestResult.java#L43)) carries `testCaseId`. Allure 3's reader maps the same field onto its identity slot. Allure 2's `Allure2Plugin#convert` never reads `result.getTestCaseId()`, so the value is silently dropped on import even though the entity has a `testId` field ready for it.

This PR maps `testCaseId` → `testId` during conversion. No new API, no UI change; the field is now reachable to widgets and downstream plugins that expect stable test identity (history, retries, baseline comparison). Tightens v2/v3 parity at zero cost.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


baev:
there is no point of such change

grootstebozewolf:
Closing per reviewer feedback: `entity.TestResult#testId` has no consumer in this codebase (no `getTestId` calls anywhere, FE doesn't read it), and stable test identity for history/retries is already carried by `historyId`, which is mapped on the line above. The PR adds a dead write. Apologies for the noise.

PR #3336, opened 2026-05-10
Preserve titlePath from on-disk schema during Allure 2 import

grootstebozewolf:
### Context

The on-disk Allure 2 schema (defined in [allure-java's `io.qameta.allure.model.TestResult`](https://github.com/allure-framework/allure-java/blob/master/allure-model/src/main/java/io/qameta/allure/model/TestResult.java#L46)) carries `titlePath`, used by Allure 3 readers to render hierarchical suite / group context. Allure 2's `Allure2Plugin#convert` never reads `result.getTitlePath()`, so the data is dropped on import and any structure downstream test frameworks have already produced becomes invisible to the v2 report.

This PR stops the data from being dropped: when non-empty, `titlePath` is stashed on the entity via the generic extra-block channel (`entity.TestResult#addExtraBlock`). This is the smallest possible preservation change — it makes the data reachable to widgets and plugins without committing to a specific UI projection.

Whether to elevate `titlePath` to a first-class field on `entity.TestResult`, render it as a breadcrumb widget, or project it onto labels is intentionally a follow-up; this PR only stops the information from being lost.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


grootstebozewolf:
Closing for the same reason as #3335 and #3338: there is no consumer of `titlePath` anywhere in the repo (Java, FE, plugins) — adding it to the extra-block channel is a dead write today. As I noted in the PR body, the actual question is the UI projection (breadcrumb / labels / first-class field), and that belongs in a discussion that produces a concrete consumer first. Apologies for the noise.

PR #3338, opened 2026-05-10
Preserve known and muted markers from on-disk statusDetails on import

grootstebozewolf:
### Context

Sister to #3336. The on-disk Allure 2 schema (allure-java's [`StatusDetails`](https://github.com/allure-framework/allure-java/blob/master/allure-model/src/main/java/io/qameta/allure/model/StatusDetails.java)) carries `known` and `muted` booleans alongside `message`, `trace`, and `flaky`. Allure 3's reader maps all five. Allure 2's `Allure2Plugin#convert` reads only the first three, so triage state is silently dropped on import.

This PR stops the data from being lost: when set, `known` and `muted` are stashed on the entity via the generic extra-block channel (`entity.TestResult#addExtraBlock`), keyed `statusDetailsKnown` / `statusDetailsMuted`. Same shape as the recent `titlePath` preservation change — minimal data preservation, no commitment to a specific UI projection.

Whether to elevate these to first-class fields on `entity.TestResult`, project as labels, or render dedicated badges is a follow-up worth its own discussion; this PR only stops the information from being lost.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/allure-generator/src/main/java/io/qameta/allure/allure2/Allure2Plugin.java
+++ b/allure-generator/src/main/java/io/qameta/allure/allure2/Allure2Plugin.java
@@ -199,6 +199,15 @@ public class Allure2Plugin implements Reader {
             dest.setStatusMessage(details.getMessage());
             dest.setStatusTrace(details.getTrace());
             dest.setFlaky(details.isFlaky());
+            // Preserve known / muted markers from the on-disk schema so triage state
+            // stays reachable on the entity (Allure 3 readers consume these fields).
+            // Stored only when set to keep the extra-block free of default-valued noise.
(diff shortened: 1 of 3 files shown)


baev:
there is no point of such change

grootstebozewolf:
Closing per reviewer feedback. Same situation as #3335: there is no consumer of `statusDetailsKnown` / `statusDetailsMuted` anywhere in the repo (no FE reader, no plugin reading the extra block), so this is a dead write. As I noted in the PR body, projecting these into a real UI surface is the open question — until that's settled, preserving them just adds noise. Apologies for the churn.

PR #3339, opened 2026-05-10
Remove deprecated ReportWebPlugin

grootstebozewolf:
### Context

`ReportWebPlugin` is annotated `@Deprecated(forRemoval = true)` and documented to be replaced by `ReportWebGenerator` (used today by `ReportGenerator`). A repo-wide search confirms zero remaining references outside its own file: no test class, no service registration, no callers in `plugins/` or any of the build scripts.

Beyond being unused, the class carries two latent bugs that would mislead anyone who did instantiate it via the deprecated `Aggregator` interface:

1. It reads `ALLURE_NO_ANALYTICS` through `Optional…map(parseBoolean)…orElse(false)`, so the value documented in [`Analytics.md`](Analytics.md) (`export ALLURE_NO_ANALYTICS=1`) does not disable analytics on this code path. (Same defect as the live path addressed in #3333.)
2. It puts the resulting flag into the FreeMarker data model under the key `Constants.NO_ANALYTICS` (`"ALLURE_NO_ANALYTICS"`), but `index.html.ftl` reads the variable named `analyticsDisable`. The opt-out is therefore a no-op regardless of the env var value.

Removing the class eliminates the duplicate-logic surface where these drift bugs could accumulate. `ReportWebGenerator` remains the single, tested path through `index.html.ftl` rendering.

Net deletion (113 lines, one file).

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests _(no test changes — existing `ReportWebGeneratorTest` covers the surviving path)_


baev:
There is no point in making such a change. Some users may still use it when choosing `allure-generator` as a dependency. The Java generator is effectively EOL, since Allure 3 can generate the same report, so it is better not to introduce unnecessary breaking changes.

grootstebozewolf:
Conceded — EOL-maintenance economics make the case against removal stronger than the case for it, and you have visibility into external `allure-generator` consumers that I don't. The latent bugs I flagged in the body existed before this PR, so removing the class doesn't help anyone using it. Thanks for the steer.

PR #3432, opened 2026-07-25
Fix plain text values being treated as relative URLs

ronitchiluka06:
## Description

Fixes an issue where plain-text parameter values were incorrectly treated as relative URLs and rendered as clickable links.

## Changes

- Removed the catch-all relative URL condition from `isRelativeUrl`.
- Plain-text values are now rendered as normal text.

## Verification

Ran:

- npm run typecheck:web
- npm run lint
- npm run build


baev:
Thanks! 

fixed via https://github.com/allure-framework/allure2/pull/3435

PR #3476, opened 2026-08-26
Support run-level errors and attachments in Allure 2 reports

baev:
### Context

This adds first-class run-level errors and attachments to Allure 2 reports. Integrations can report setup and container failures without creating fake test results, keeping test statistics accurate while clearly marking the overall run unsuccessful—even when every reported test passes or a fixture failure prevents tests from being reported.

Reports with run-level data now show a prominent warning on Overview and a contextual Run tab with error details, traces, expected and actual values, and attachment previews and downloads. The Run tab remains hidden when no global errors or attachments exist, preserving the navigation of existing reports.

Screenshot 2026-08-26 at 20 20 24

Screenshot 2026-08-26 at 20 20 34

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/allure-plugin-api/src/main/java/io/qameta/allure/entity/GlobalAttachment.java
+++ b/allure-plugin-api/src/main/java/io/qameta/allure/entity/GlobalAttachment.java
@@ -29,6 +29,7 @@ public class GlobalAttachment implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    protected Long timestamp;
     protected String uid;
     protected String name;
     protected String source;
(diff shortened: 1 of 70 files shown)


allure-plugin-api/src/main/java/io/qameta/allure/entity/GlobalAttachment.java
formaceft-93, line 1: Global attachments in the current Allure globals model carry `timestamp` as well as global errors. This entity drops it, so `widgets/globals.json` loses attachment chronology. Please add `Long timestamp` here and map it from `Allure2Globals.Attachment` in `Allure2GlobalsReader`; also update the frontend `GlobalAttachment` type and a regression assertion.

formaceft-93 (approved):

formaceft-93 (approved):

